PR #1, opened 2019-10-09
Migrated use of CloudFlare font awesome to local paths to remove netw…

Guydalf:
Migrated use of Cloud Flare font awesome to local paths to remove network dependency


scovetta:
I think we should do this a different way. Rather than embedding the resources as source code, we should consider using Yarn or something like that, and include a manifest file that describes the JavaScript to download during build. That way (a) our dependencies are discoverable; (b) we have a way of easily updating versions; and (c) we keep our repo to "just our code" as much as possible.

End result should be the same -- after build, we'll have local resources.

Longer term, it may make sense to consider just running a local web-server for the content and deprecate the local HTML page. I'll open an issue to keep this on our radar.

Guydalf:
Great idea.  I had worried about getting out of sync with dependency versions with fixes and more.

Guydalf:
I'm going to go ahead and merge this for now until it is reworked to keep aligned with what people were told i.e. that it does not require a network connection to run.  Then will replace it.

PR #13, opened 2019-10-21
Addresses feature issue for showing confidence level and restoring dy…

Guydalf:
…namic grouping to match html/json using liquid (not c# text replace) for best of both worlds solution.


Guydalf:
#13

Issue #5, opened 2019-10-11
Match details exclusion from json


Guydalf:
**Describe the bug**
With a recent change over in the html report we can no longer have an indicator for a given feature tag in the html report without all instances appearing as match records also I the json output.  Fix the issue so that the two possible so we don't have multiple records or noise for things we don't want to see e.g. ability to list dependencies I the html and aggregated sections of each report without a large list of matches in the json output.

Guydalf:
this is essentially fixed now but needs the exclusion list added by to the rulesprocessor constructor

PR #20, opened 2019-10-24
Restores feature allowing count of specified tags that are not wanted…

Guydalf:
… in the match details output i.e. classes, functions, exceptions, html forms, comments...Also splits general comments from fix comments.


Guydalf:
Fixes #5

PR #37, opened 2019-11-20
Several key rules/patterns added to address missing auth and auth pat…

Guydalf:
Several key rules/patterns added to address missing auth and auth patterns for SAML, Oauth, Basic Auth, a missing Microsoft login URL, active directory, LDAP, NTML, and more.  Also corrected few patterns with regex type with '.' but no literal markers.  Some work done on infra file.


Guydalf:
Moving forward as Mike is busy and request was for FYI primarily.

PR #48, opened 2019-12-07
Fix for final exception handling in cases where log file is not yet c…

Guydalf:
…reated successfully.  Also updates metric for URL counts and test -x argument values.


Guydalf:
One note: removed debug based exception behavior to show stack to test with release behavior.

Issue #97, opened 2020-01-23
tagdiff command Runtime error: Unexpected character encountered


imrzegarra:
**Describe the bug**
dotnet AppInspector.dll tagdiff --src1 /media/emailba1 --src2 /media/emailba2

2020-01-23T14:45:21 1 ERROR - A runtime error has occured.  Please see log file for more information.
2020-01-23T14:45:21 1 ERROR - Runtime error: Unexpected character encountered while parsing value: M. Path '', line 0, position 0.    at Newtonsoft.Json.JsonTextReader.ParseValue()
   at Newtonsoft.Json.JsonReader.ReadAndMoveToContent()
   at Newtonsoft.Json.JsonReader.ReadForType(JsonContract contract, Boolean hasConverter)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)
   at Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader reader, Type objectType)
   at Newtonsoft.Json.JsonSerializer.Deserialize(JsonReader reader, Type objectType)
   at Newtonsoft.Json.JsonConvert.DeserializeObject(String value, Type type, JsonSerializerSettings settings)
   at Newtonsoft.Json.JsonConvert.DeserializeObject[T](String value, JsonSerializerSettings settings)
   at Newtonsoft.Json.JsonConvert.DeserializeObject[T](String value)
   at Microsoft.AppInspector.TagDiffCommand.Run()
   at Microsoft.AppInspector.Program.RunTagDiffCommand(TagDiffCommandOptions opts)
   at Microsoft.AppInspector.Program.<>c.<Main>b__6_1(TagDiffCommandOptions opts)
   at CommandLine.ParserResultExtensions.MapResult[T1,T2,T3,T4,T5,TResult](ParserResult`1 result, Func`2 parsedFunc1, Func`2 parsedFunc2, Func`2 parsedFunc3, Func`2 parsedFunc4, Func`2 parsedFunc5, Func`2 notParsedFunc)
   at Microsoft.AppInspector.Program.Main(String[] args)

**To Reproduce**
Execute 
dotnet AppInspector.dll tagdiff --src1 /media/emailba1 --src2 /media/emailba2

**Expected behavior**
End The command Succesfully and generates output.html

**Desktop (please complete the following information):**
 - OS: Debian Linux
 - Version 5.2

[...]

PR #98, opened 2020-01-24
Removed ComVisible pattern which can be a false positive for code bei…

guyacosta:
Removed ComVisible pattern which can be a false positive for code being a COM object.  Restores clarity to x509Certificate tag that was lost at some point earlier.  Adds additional patterns for web.service identification; misc pattern improvement.  Fixes issue with TagDiff and TagTest commands.


guyacosta:
Fix for #97

Issue #125, opened 2020-01-31
JSON output should be pipe-able


scovetta:
**Describe the bug**
If you output JSON to the console, any non-JSON output should be sent either to stderr, or suppressed.

```
PS C:\ai> .\AppInspector.exe analyze -s D:\src\npmjs.com\m\md\MD5@1.0.0.tgz -t -f json
Microsoft Application Inspector 1.0.24
Analyze command running
Decompressing files...
100% source files processed
Preparing report
[
  {
    "tags": [
      "Cryptography.Encryption.General",
      "Cryptography.HashAlgorithm.Legacy",
      "Metadata.Application.Author",
      "Metadata.Application.Description",
      "Metadata.Application.Version",
      "Metadata.CloudServices.Code.Repo.GitHub"
    ]
  }
]
Analyze command completed
```

It would be useful to be able to pipe this to jq, ConvertFrom-Json, or another processor, but currently the JSON output and the logs are sent to stdout so can't be easily filtered.

**Expected behavior**
When output type is text or JSON, anything that isn't core output should be sent to stderr instead.

guyacosta:
Fixed.

PR #126, opened 2020-02-01
Removes a string from the output that was affecting the json output f…

guyacosta:
…ormat and consequently a prior workaround fix for the same issue.


guyacosta:
fix for #125

Issue #109, opened 2020-01-27
Report Show Skipped Code as Analyzed


imrzegarra:
**Describe the bug**
The final report output.html show analyzed 0.99% but detects several kinds of code, I think the result is wrong and this could be the skipped %.

**Desktop (please complete the following information):**
 - OS: Windows 10 - Linux 
 - Browser FireFox
<img width="271" alt="Captura de Pantalla 2020-01-27 a la(s) 08 55 57" src="https://user-images.githubusercontent.com/60227319/73173091-4eadfa80-40e3-11ea-98df-08319f96e7e0.png">
<img width="834" alt="Captura de Pantalla 2020-01-27 a la(s) 08 58 09" src="https://user-images.githubusercontent.com/60227319/73173092-4eadfa80-40e3-11ea-97f6-84b6f99bb62b.png">

guyacosta:
The files analyzed is simply the number of files that met the language criteria while the number skipped are those that did not e.g. media files of some type or a programming language we didn't recognize yet.  The percentage completed represents total files considered including those that were skipped and those actually analyzed divided by the total files enumerated so that 100% means everything we found not 100% of just those analyzed but the analysis at a high level included determining we don't support feature analysis of some files.  

 int totalFilesReviewed = _appProfile.MetaData.FilesAnalyzed + _appProfile.MetaData.FilesSkipped;
 int percentCompleted = (int)((float)totalFilesReviewed / (float)_appProfile.MetaData.TotalFiles * 100);

The report identification of languages is from the list of those not skipped or those that we support scanning while the skipped files may include languages not yet supported.  Hope that helps.

imrzegarra:
Hi Guy, The question is why with the same code analyzed in the release 10.0.21 the number of languagues detected was less than the new release 10.0.24 but the skipped files increase. I think it should be the oposite. With more languagues less skipped.
10.0.21 (pam.xml - java)
10.0.24 (pam.xml-java-json-html)
Thanks

guyacosta:
The additional information is helpful.  I'll investigate further.

guyacosta:
The under reporting in languages earlier was an error not related to skipped files.  Several changes about the same time had been made i.e.

[...]

PR #127, opened 2020-02-02
Minor language detection fix; consistent order checks for zip/non zip…

guyacosta:
… file processing.


guyacosta:
Related in part to #109

PR #148, opened 2020-03-02
Fixed typo in database rules

PabloClon:


guyacosta:
not a typo as the old value was good but the addition discrimination on a word boundary is valid.  thanks.

Issue #163, opened 2020-03-10
NuGet Command Pkg Install Doesn't Copy Default Rules to Local App


guyacosta:
**Describe the bug**
After installing the Microsoft.CST.ApplicationInspector.Command pkg and running the containing app, the calls fail due to the default rules folder missing from where the dll is copied to i.e. the containing app \bin\debug or \bin\release folders causing an exception.  A default or custom rules path is expected.  The caller wanting to use the default ruleset would have to copy them manually from the nuget install location locally into their local app bin folder

**To Reproduce**
1. Download the pkg from Nuget via Visual Studio into an app
2. Create a AnalyzeCommandOption object
3. Create an AnalyzeCommand object
4. Call GetResults method on the command object.  See the exception from step 3.

**Expected behavior**
User does not have to manually copy over the Rules folder into their bin folders for use.

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

guyacosta:
@daalcant let me know if a code change is required apart from the nuget packaging to resolve.  I'm looking into also.

gfs:
My recommendation is to have default rules be built and accessed as embedded resources so you don't have to include separate files with the library.

-------- Original Message --------

gfs:
From [Attack Surface Analyzer](https://github.com/microsoft/AttackSurfaceAnalyzer/blob/master/Lib/Utils/Analyzer.cs) this is how I load the default rules and analysis json files.

```
            try
            {
                var assembly = typeof(FileSystemObject).Assembly;
                var resourceName = "AttackSurfaceAnalyzer.analyses.json";

                using (Stream stream = assembly.GetManifestResourceStream(resourceName))
                using (StreamReader reader = new StreamReader(stream))
                {
                    config = JsonSerializer.Deserialize<RuleFile>(reader.ReadToEnd());
                    Log.Information(Strings.Get("LoadedAnalyses"), "Embedded");
                }
                if (config == null)
                {
                    Log.Debug("No filters today.");
                    return;
                }
            }
```

gfs:
Since your rules are formatted similar to DevSkim we can pack them before inclusion using something like we do in the devskim csproj.

```
  <Target Name="PreBuild" BeforeTargets="BeforeCompile">
    <Exec Command="dotnet ../../RulePacker/devskim.dll pack ../../rules Resources/devskim-rules.json" />
  </Target>
```

[...]

PR #165, opened 2020-03-11
Fix for #163

guyacosta:


guyacosta:
p.s. Since the PackRules command automatically does rule verification there is no need to support checking default rules as part of the VerifyRules command so I removed that leaving it focused on checking user custom rules only hence a could of the argument changes you may notice.  Also, the TagDiff and TagTest commands do not need to specify default args that are already set by default now so misc stuff but primary purpose is for #163

gfs:
You should take a look at what we do in DevSkim to pack the rules as part of the build itself (see my note above).  That means you won't have to manually remember to pack them, and the pipeline and your machine will build them the same way.  Let me know if you need a hand with that.  @daalcant might also be able to help.  I highly recommend not relying on manually remembering to pack the rules before checking them in.

guyacosta:
> You should take a look at what we do in DevSkim to pack the rules as part of the build itself (see my note above). That means you won't have to manually remember to pack them, and the pipeline and your machine will build them the same way. Let me know if you need a hand with that. @daalcant might also be able to help. I highly recommend not relying on manually remembering to pack the rules before checking them in.

I'll give that another look -we'll achieve that to enhance.  I thought we might have a chicken before the egg but I'll review again.  I'll be in touch by early afternoon if stuck.

guyacosta:
Alright dudes it works.  Thanks to @gfs for the suggested solution -very clever.  @daalcant let's do have your final review as well and call it done.

guyacosta:
Hmm @daalcant seeing a new error on the checks -doesn't look related to my change.  Please assist.

AppInspector.sln
```diff
@@ -16,6 +16,26 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MultiExtractor", "MultiExtr
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AppInspector.CLI", "AppInspector.CLI\AppInspector.CLI.csproj", "{824ED27E-A4CF-46A6-A01F-98B0821EB61C}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "RulesPacker", "RulesPacker", "{C464D0CE-5254-4EA5-87C9-C0C96E40C3CB}"
+	ProjectSection(SolutionItems) = preProject
+		RulesPacker\ApplicationInspector.CLI.deps.json = RulesPacker\ApplicationInspector.CLI.deps.json
+		RulesPacker\ApplicationInspector.CLI.dll = RulesPacker\ApplicationInspector.CLI.dll
+		RulesPacker\ApplicationInspector.CLI.exe = RulesPacker\ApplicationInspector.CLI.exe
+		RulesPacker\ApplicationInspector.CLI.pdb = RulesPacker\ApplicationInspector.CLI.pdb
```
gfs: You don't need the symbols (.pdb files)

AppInspector.sln
```diff
@@ -16,6 +16,26 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MultiExtractor", "MultiExtr
...
+		RulesPacker\ApplicationInspector.CLI.runtimeconfig.dev.json = RulesPacker\ApplicationInspector.CLI.runtimeconfig.dev.json
```
gfs: I don't think you need this dev config json.

AppInspector/AppInspector.Commands.csproj
```diff
@@ -84,9 +82,12 @@
...
   <ItemGroup>
-    <None Include="..\LICENSE.txt" Pack="true" PackagePath=""/>
-    <None Include="..\icon-128.png" Pack="true" PackagePath=""/>
-    <None Update="rules\**" Pack="true" PackagePath="lib\$(TargetFramework)\"/>
+    <None Include="..\LICENSE.txt" Pack="true" PackagePath="" />
+    <None Include="..\icon-128.png" Pack="true" PackagePath="" />
   </ItemGroup>
 
+  <Target Name="PreBuild" BeforeTargets="BeforeCompile">
+    <Exec Command="..\RulesPacker\ApplicationInspector.CLI.exe packrules -r rules\default -o Resources\defaultRules.json" />
```
gfs: use `dotnet ../RulesPacker/ApplicationInspector.CLI.dll` as it will be cross platform

AppInspector/Commands/TagTestCommand.cs
```diff
@@ -23,7 +23,6 @@ enum TagTestType { RulesPresent, RulesNotPresent }
         private string _arg_srcPath;
         private string _arg_customRulesPath;
         private string _arg_outputFile;
-        private bool _arg_ignoreDefaultRules;
```
gfs: Is there still value in supporting suppressing the default rules here? Should still be possible, just requires dumping out the embedded filters if the user requests it.

AppInspector/Resources/defaultRules.json
```diff
@@ -0,0 +1,17075 @@
+[
```
gfs: Delete this entire file, it isn't needed as it'll get generated dynamically.  You can add `Resources/defaultRules.json` to the `.gitignore` to make sure no one checks it in by mistake.

RulesEngine/PatternType.cs
```diff
@@ -34,7 +34,7 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
                     break;
             }
             writer.WriteValue(svrstr);
-            writer.WriteValue(svr.ToString().ToLower());
+            //writer.WriteValue(svr.ToString().ToLower());
```
gfs: Run codemaid.

AppInspector/AppInspector.Commands.csproj
gfs: This from DevSkim works on Windows and Mac

```
<Target Name="PreBuild" BeforeTargets="BeforeCompile">
    <Exec Command="dotnet ../../RulePacker/devskim.dll pack ../../rules Resources/devskim-rules.json" />
  </Target>
```
```suggestion
    <Exec Command="dotnet ../RulesPacker/ApplicationInspector.CLI.exe packrules -r rules/default -o Resources/defaultRules.json" />
```

gfs: This could be what is breaking the linux/mac build.

gfs:
Also exclude the .exe - its not necessary, you can `dotnet` run the dll directly.

AppInspector.sln
guyacosta: right

AppInspector/AppInspector.Commands.csproj
guyacosta: it complains it can't find hostpolicy.dll when I use the dll method but yes, makes sense for other OS's.  I'll review to address.

guyacosta: also using the / stead of \ causes the build to fail as it can't locate the file i.e. error 9009 when building on windows.  At this point I welcome some help to short cut the final commit to address if you're interested.

AppInspector/Commands/TagTestCommand.cs
guyacosta: no there is no value in testing against the default set only a custom set.  testing against the default set is equivalent to just running analyze command which is why I removed it.

AppInspector/Resources/defaultRules.json
guyacosta: Yep, I see.

RulesEngine/PatternType.cs
guyacosta: Have Code Cleanup On Save.  I'll remove manually.

AppInspector/AppInspector.Commands.csproj
gfs: I'm checking out the code now.

gfs:
Checking windows now.

gfs:
I think its a pipeline issue at this point - that still stops us from merging though because we won't be able to publish. I can build and run this fine on both Windows and Mac now. @daalcant can you take a look?

guyacosta:
I've got it fixed.  Standby.  New commit coming.

AppInspector/AppInspector.Commands.csproj
guyacosta: Ok.  I did get it working locally complete with \ and dll but I see you committed a change that is similar.  I had also removed the unnecessary files including the defaultRules.json.  Thanks!  We'll go with the changes you made.

guyacosta:
Repeat from above for my ending comment: Ok. I did get it working locally complete with \ and dll but I see you committed a change that is similar. I had also removed the unnecessary files including the defaultRules.json. Thanks! We'll go with the changes you made.  If we're still having pipeline issues, will allow @daalcant to review for nuget or other related issues.  Thanks.

daalcant:
Checking this out. I'll add a pipeline fix to this branch.

daalcant:
I believe it's just an issue with working directories. Will make some changes and test it out, should be an easy fix.

daalcant:
Scratch that, not a working dir issue. 

I can sporadically get the same pipeline error on my local builds. I think this is a race condition that we also saw with DevSkim, where the rule packing step is being executed twice (once for the CLI, once for the library) and locking itself out when trying to write the rules file.

gfs:
The CLI doesn't have a rules packing step in its csproj, it should only be being triggered on the lib though.

daalcant:
> 
> 
> The CLI doesn't have a rules packing step in its csproj, it should only be being triggered on the lib though.

Correct, but the lib dependency causes it to run again.

daalcant:
Troubleshooting this to see if there's a better prebuild event to target, or if we can perhaps find a way to ensure the rule packing is only executed once.

daalcant:
Identified that this behavior only occurs when using the "publish" command from the solution directory. I can work around it easily in the pipeline, but going to see if there's a "better" way before falling back to that.

guyacosta:
Ok.  If we have we can separate out the packing operation but is the model being used is the same as DevSkim which isn't exhibiting the issue?  Ping me for more immediate response if needed.  Thx.

gfs:
Per Davids research we believe this could occur in DevSkim we just hadn't encountered this. We should have a fix/workaround until they fix what appears to be an issue in .NET Core

daalcant:
Potentially related issue is: https://github.com/dotnet/sdk/issues/3707

Workarounds via csproj targets don't seem to be successful, so I will modify the pipeline to publish from the CLI directory which prevents the duplicate pre-build event.

guyacosta:
Nice that should fix it because the CLI is built last.

daalcant:
Looks like the pipeline fix was successful. I'm going to verify all of the produced artifacts to make sure I didn't overlook anything before we merge.

guyacosta:
Nice.  I'll get the branch and perform local testing once more also before merge.

guyacosta:
Thanks.  All tests have passed.

Issue #179, opened 2020-03-25
AppInspector overrides logging level for all loggers


scovetta:
**Describe the bug**
In [Utils.cs](https://github.com/microsoft/ApplicationInspector/blob/b827a8297d5659af9b8c9dbc08eb5d0690418fa8/AppInspector/Utils.cs#L270), the effect of setting the LoggingLevel for "*" seems to overwrite any other classes using NLog for logging:

```
            {
                config.AddTarget(fileTarget);
                config.LoggingRules.Add(new LoggingRule("*", log_level, fileTarget));
            }

            LogFilePath = opts.LogFilePath;//preserve for console path msg

            LogManager.Configuration = config;
```

I think the LoggingRule should only be set for Application Inspector namespaces, and the target's name should be more specific to avoid colliding with an existing target.

The result of this is that once AppInspector is called with LogFileLevel == "Off", all other loggers fail to produce any output.

guyacosta:
ouch -all right.  Thanks for reporting.  Will get addressed asap.

scovetta:
Re-opening, this is overriding previously created loggers:

https://github.com/microsoft/ApplicationInspector/blob/master/AppInspector/Utils.cs#L181
https://github.com/microsoft/ApplicationInspector/blob/master/AppInspector/Utils.cs#L264

```
var config = new NLog.Config.LoggingConfiguration();
...
LogManager.Configuration = config;
```

This is creating a new log configuration, assigning the new target and config, which all looks fine, but then you overwrite the global LogManager.Configuration with the one you created.

What you need to do (I think) is:

```
var config = NLog.LogManager.Configuration;
...
LogManager.Configuration = config;
```

(I'd also suggest you change the log name from `CST.ApplicationInspector` to `Microsoft.CST.ApplicationInspector`.)

guyacosta:
With minor change the new check provides the flexibility to default to a new config file as before if one has not been defined in the global NLog space, reuse an existing one as well as simply take a caller defined log as an argument complete with defined configuration which was the original intent.  All three options work now so you can have separate targets, as combined target and layout etc.  

Final check was to add this:
            LoggingConfiguration config = LogManager.Configuration;
            if (config == null)//fix #179 to prevent overwrite of caller config...i.e. just add ours
            {
                config = new LoggingConfiguration();
            }

PR #180, opened 2020-03-26
Fix for #179

guyacosta:


guyacosta:
Adds named Logger to enable other uses of NLog from calling apps

PR #189, opened 2020-03-30
Major add of unit tests for most commands.  Some fixes resulting...

guyacosta:
Major add of unit tests for most commands.  Some fixes as a resulting benefit of these new tests to improve the core functionality and handle outlier behavior.


guyacosta:
Adds missing options for fileexlusionfilters to TagDiff and TagTest commands preventing ability to use with files under test, docs etc. folders
Adds check for ability to write log file where requested
Adds check for consoleverbosity = none and no output file specified for all i.e. no output anywhere
Standardizes exit code names between tagdiff and tagtest; no values change
Lowercases json tags fields for conformity
Addresses count error in total matches by deducting tagcounters in preferences with includeAsMatch=false
Adds unit tests and source files (early draft for now)

PR #192, opened 2020-04-06
1.2 dev branch

### LINKED ISSUE #191: Return result objects not json/text for NuGet calls
The initial rollout of NuGet support essentially wrapped the CLI functionality which supports writing to a file or the console as output.  This is not the final state which is the return fully formed result objects that contain the meta-data and detailed matches.

**Enhancements**:
Provide a result object for calls into the Application.Commands object
Provide options that are more applicable for DLL use i.e. don't take file output paths or console arguments etc. which should only be applicable for CLI use.

guyacosta:
Unit tests Module
•	189 passing tests added covering both NuGet and CLI entry points under separate namespaces/classes
•	Improved overall code quality and reporting of invalid args/combinations

Overall Code Structure
•	Refactoring+rework for technical debt getting to 1.0 eliminating unused code, renaming, moving to isolate to right project
•	*Object results-based support for NuGet callers rather than json/text strings
•	*New support for First/Best match options for results (best match elevates higher confidence value matches)
•	Separation of CLI vs Operation argument options
•	All file output format and writters moved to CLI project
•	Cleaner isolation of functionality and readability
•	Text and Json output support added for all commands (previously limited to text for all but analyze)
•	Better use of inheritance to organize common arguments and override 
•	Better use of polymorphism for checking arguments 

Resolves #191


guyacosta:
Suggest to download and review PR branch in it's entirely due to the size of change.

Notes:
Detailed Code Changes
CLI Module
•	All file writers moved to this project out of command/core module that return objects only
•	All command options include output args
•	All common options in base class
•	Check for output file path write before running op
•	Checks added for log write ability to specified path before operation starts
Command Module
•	All arguments exclude CLI only output options 
•	All write statements related to output removed = no IO other than log
•	GetResult() method for each command returning Results object not json string
•	Result objects include standard result code and embedded objects that are serializable and only contain methods specific to the result data including the AppProfile class which is gone now
•	Renaming of classes, properties, variables for readabilty, accuracy e.g. MsgHelp, not ErrMsg, ScanResult not Issue in RulesEngine
•	Tag counter support outside of CLI entry point; no dependency on preferences using "Metric" tag type as key
RulesEngine Module
•	*Added support for First vs Best match (perf/quality) with AllowDuplicates option
Security
•	Check for log path being used to overwrite non app log file

AppInspector.CLI/CLICmdOptions.cs
```diff
@@ -2,46 +2,42 @@
...
-
-        [Option('e', "text-format", Required = false, HelpText = "Match text format specifiers", Default = "Tag:%T,Rule:%N,Ruleid:%R,Confidence:%X,File:%F,Sourcetype:%t,Line:%L,Sample:%m")]
-        public string TextOutputFormat { get; set; }
-
         [Option('r', "custom-rules-path", Required = false, HelpText = "Custom rules file or directory path")]
         public string CustomRulesPath { get; set; }
 
-        [Option('t', "tag-output-only", Required = false, HelpText = "Output only identified tags", Default = false)]
-        public bool SimpleTagsOnly { get; set; }
+        [Option('h', "match-depth", Required = false, HelpText = "First match or best match based on confidence level (first|best)", Default = "best")]
```
gfs, line 39: This should be `string?`.

Only marking one of these.  https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/nullable-reference-types

AppInspector.CLI/CLICmdOptions.cs
```diff
@@ -58,14 +54,29 @@ public class AnalyzeCommandOptions : AllCommandOptions
...
+
+        [Option('t', "tag-output-only", Required = false, HelpText = "Output only identified tags", Default = false)]
+        public bool SimpleTagsOnly { get; set; }
+
+        public CLIAnalyzeCmdOptions()
+        {
+            OutputFileFormat = "html";
+            MatchDepth = "best";
+            IgnoreDefaultRules = false;
+            SimpleTagsOnly = false;
```
gfs, line 71: Why set this default value twice?  See line 63.

AppInspector.CLI/ResultsWriter.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// <param name="appProfile"></param>
+        /// <param name="options"></param>
+        public static void Write(AnalyzeResult analyzeResult, CLIAnalyzeCmdOptions options)
+        {
+            int MAX_HTML_REPORT_FILE_SIZE = 1024 * 1000 * 3;  //warn about potential slow rendering
+
+            //analyze with html format limit checks
+            if (options.OutputFileFormat == "html")
+            {
+                options.OutputFilePath = "output.html";
```
gfs, line 64: What if the user set an output file path?

AppInspector.CLI/CLICmdOptions.cs
```diff
@@ -2,46 +2,42 @@
...
+        [Option('f', "output-file-format", Required = false, HelpText = "Output format [json|text]", Default = "text")]
+        public string OutputFileFormat { get; set; }
 
-        [Option('l', "log-file-path", Required = false, HelpText = "Log file path")]
-        public string LogFilePath { get; set; }
-        [Option('v', "log-file-level", Required = false, HelpText = "Log file level [Debug|Info|Warn|Error|Trace|Fatal|Off]", Default = "Error")]
-        public string LogFileLevel { get; set; }
+        public CLICommandOptions()
+        {
+            OutputFileFormat = "text";
```
gfs, line 22: I would recommend built in default declarations functionality in the option definition or even `{ get; set; } = value` the property declaration over hardcoded values in the empty `.ctor`

AppInspector.CLI/Program.cs
```diff
@@ -25,103 +26,364 @@ static int Main(string[] args)
...
+                    CLIVerifyRulesCmdOptions,
+                    CLIPackRulesCmdOptions>(args)
                   .MapResult(
-                    (AnalyzeCommandOptions opts) => RunAnalyzeCommand(opts),
-                    (TagDiffCommandOptions opts) => RunTagDiffCommand(opts),
-                    (TagTestCommandOptions opts) => RunTagTestCommand(opts),
-                    (ExportTagsCommandOptions opts) => RunExportTagsCommand(opts),
-                    (VerifyRulesCommandOptions opts) => RunVerifyRulesCommand(opts),
-                    (PackRulesCommandOptions opts) => RunPackRulesCommand(opts),
+                    (CLIAnalyzeCmdOptions cliOptions) => VerifyOutputArgsRun(cliOptions),
```
gfs, line 36: Why do you have separate options for the CLI and the Library?  It looks like they overlap nearly entirely.

AppInspector.CLI/Program.cs
```diff
@@ -25,103 +26,364 @@ static int Main(string[] args)
...
+            List<string> validFormats = new List<string>();
+            validFormats.Add("json");
+            validFormats.Add("text");
+
+            if (options is CLIAnalyzeCmdOptions)
+            {
+                fileFormatArgs = ((CLIAnalyzeCmdOptions)options).OutputFileFormat;
+            }
+
+            bool isValidFormat = validFormats.Any(v => v.Contains(fileFormatArgs));
```
gfs, line 209: This should probably be .Equals, not .Contains

AppInspector.CLI/Program.cs
```diff
@@ -25,103 +26,364 @@ static int Main(string[] args)
...
+        /// <param name="options"></param>
+        static void CommonOutputChecks(CLICommandOptions options)
+        {
+            //validate requested format
+            string fileFormatArgs = options.OutputFileFormat;
+            List<string> validFormats = new List<string>();
+            validFormats.Add("json");
+            validFormats.Add("text");
+
+            if (options is CLIAnalyzeCmdOptions)
```
gfs, line 204:
```suggestion
            if (options is CLIAnalyzeCmdOptions cliAnalyzeOptions)
```

AppInspector.CLI/Program.cs
```diff
@@ -25,103 +26,364 @@ static int Main(string[] args)
...
+            {
+                fileFormatArgs = ((CLIAnalyzeCmdOptions)options).OutputFileFormat;
```
gfs, line 206:
```suggestion
                fileFormatArgs = cliAnalyzeOptions.OutputFileFormat;
```

AppInspector.CLI/Program.cs
```diff
@@ -25,103 +26,364 @@ static int Main(string[] args)
...
+                File.WriteAllText(filePath, "");//verify ability to write to location
+            }
+            catch (Exception)
+            {
+                WriteOnce.Error(MsgHelp.FormatString(MsgHelp.ID.CMD_INVALID_FILE_OR_DIR, filePath));
+                throw new OpException(MsgHelp.FormatString(MsgHelp.ID.CMD_INVALID_FILE_OR_DIR, filePath));
+            }
+
+        }
+
```
gfs, line 253: This is a lot of extra whitespace.

AppInspector.CLI/Program.cs
```diff
@@ -25,103 +26,364 @@ static int Main(string[] args)
...
+            }); ;
+
+            AnalyzeResult analyzeResult = command.GetResult();
+            exitCode = analyzeResult.ResultCode;
+            ResultsWriter.Write(analyzeResult, cliOptions);
+
+            return (int)exitCode;
+        }
+
+
```
gfs, line 284: Whitespace.

AppInspector.CLI/ResultsWriter.cs
```diff
@@ -0,0 +1,108 @@
...
+
+                //html report size warning
+                if (File.Exists(options.OutputFilePath) && new FileInfo(options.OutputFilePath).Length > MAX_HTML_REPORT_FILE_SIZE)
+                {
+                    WriteOnce.Info(MsgHelp.GetString(MsgHelp.ID.ANALYZE_REPORTSIZE_WARN));
+                }
+
+                WriteOnce.Operation(MsgHelp.FormatString(MsgHelp.ID.CMD_COMPLETED, "Analyze"));
+
+                if (options.OutputFileFormat == "html" && !options.SuppressBrowserOpen)
```
gfs, line 77: Why do you check if its equal to html again?

AppInspector.CLI/ResultsWriter.cs
```diff
@@ -0,0 +1,108 @@
...
+
+namespace Microsoft.ApplicationInspector.CLI
+{
+    /// <summary>
+    /// Wrapper for CLI only output arg validation which may be unique to a command 
+    /// and for allocating the correct writter type and format writter object 
+    /// </summary>
+    public static class ResultsWriter
+    {
+        public static void Write(TagDiffResult tagDiffResult, CLITagDiffCmdOptions options)
```
gfs, line 16: Looking at the writeresults functions it doesnt look like you need all these different wrappers and you could use a single wrapper with the parent classes as the arguments.

AppInspector.CLI/Writers/DummyWriters.cs
```diff
@@ -0,0 +1,88 @@
+﻿// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT License. See LICENSE.txt in the project root for license information.
+
+using Microsoft.ApplicationInspector.Commands;
+
+namespace Microsoft.ApplicationInspector.CLI
+{
+    public class AnalyzeDummyWriter : CommandResultsWriter
```
gfs, line 8: Why do you need so many of these subclasses? Nothing about them is unique.

AppInspector.CLI/Writers/VerifyRulesJsonWriter.cs
```diff
@@ -0,0 +1,28 @@
+﻿using Microsoft.ApplicationInspector.Commands;
+using Newtonsoft.Json;
+
+namespace Microsoft.ApplicationInspector.CLI.Writers
+{
+    class VerifyRulesJsonWriter : CommandResultsWriter
+    {
+        public override void WriteResults(Result result, CLICommandOptions commandOptions, bool autoClose = true)
```
gfs, line 8: This looks like the same as in TagDiffJsonWriter, could this be the definition of WriteResults in the base class and then only override it when you have something different to do?

AppInspector.CLI/Writers/VerifyRulesJsonWriter.cs
```diff
@@ -0,0 +1,28 @@
...
+            WriteOnce.NewLine();
+
+            if (autoClose)
+                FlushAndClose();
+        }
+
+        public override void FlushAndClose()
+        {
+            TextWriter.Flush();
+            TextWriter.Close();
```
gfs, line 27: You're not setting WriteOnce.TextWriter to null here but you are in other TextWriters

AppInspector.CLI/Writers/VerifyRulesJsonWriter.cs
```diff
@@ -0,0 +1,28 @@
...
+            jsonSerializer.Formatting = Formatting.Indented;
+            jsonSerializer.Serialize(TextWriter, (VerifyRulesResult)result);
+
+            WriteOnce.NewLine();
+
+            if (autoClose)
+                FlushAndClose();
+        }
+
+        public override void FlushAndClose()
```
gfs, line 24: You seem to do the same thing in all of them, do you need to override this? Could it just be part of the base class?

AppInspector.CLI/Writers/WriterFactory.cs
```diff
@@ -0,0 +1,250 @@
...
+            writer.OutputFileName = cLIAnalyzeCmdOptions.OutputFilePath;
+            writer.TextWriter = GetTextWriter(writer.OutputFileName);
+
+            return writer;
+        }
+
+
+        static CommandResultsWriter GetVerifyRulesWriter(CLICommandOptions options)
+        {
+            CLIVerifyRulesCmdOptions cLIAnalyzeCmdOptions = (CLIVerifyRulesCmdOptions)options;
```
gfs, line 181: If the first thing you do is cast this should be part of the method declaration. Otherwise this could throw an error if you give it the wrong type of command options.

AppInspector.CLI/html/resources/js/appinspector.js
```diff
@@ -50,9 +50,9 @@
 
 class TemplateInsertion {
     constructor(data) {
-        this.ap = data.AppProfile;
-        this.mt = this.ap.MetaData;
-        this.md = this.ap.FormattedMatchList;
+        this.mt = data.MetaData;
+        this.md = this.mt.detailedMatchList;
+        //this.tagCounters = this.mt.tagCounters;
```
gfs, line 55: Try not to merge in commented out code.

RulesEngine/RuleProcessor.cs
```diff
@@ -164,24 +150,41 @@ public Issue[] Analyze(string text, LanguageInfo languageInfo)
...
+                                        }
+                                    }
+                                }
+
+                                if (passedConditions)
+                                {
+                                    matchList.Add(newMatch);
+                                }
 
                                 AddRuleTagHashes(rule.Tags);
```
gfs, line 191: Should the tags be added if the match didn't pass conditions?

RulesEngine/RuleProcessor.cs
```diff
@@ -224,34 +226,96 @@ public Issue[] Analyze(string text, LanguageInfo languageInfo)
...
-            foreach (Issue m in resultsList)
+            if (_checkForOverRides)
             {
-                if (m.Rule.Overrides != null && m.Rule.Overrides.Length > 0)
+                // Deal with overrides 
+                List<ScanResult> removes = new List<ScanResult>();
+                foreach (ScanResult m in resultsList)
                 {
-                    foreach (string ovrd in m.Rule.Overrides)
+                    if (m.Rule.Overrides != null && m.Rule.Overrides.Length > 0)
```
gfs, line 252:
```suggestion
                    if (m.Rule.Overrides is List<string> overrides)
```

RulesEngine/RuleProcessor.cs
```diff
@@ -224,34 +226,96 @@ public Issue[] Analyze(string text, LanguageInfo languageInfo)
...
                     {
-                        // Find all overriden rules and mark them for removal from issues list   
-                        foreach (Issue om in resultsList.FindAll(x => x.Rule.Id == ovrd))
+                        foreach (string ovrd in m.Rule.Overrides)
```
gfs, line 254:
```suggestion
                        foreach (string ovrd in overrides)
```

RulesPacker/saveold/ApplicationInspector.CLI.deps.json
```diff
@@ -0,0 +1,209 @@
+{
```
gfs, line 1: This path looks like something that shouldn't be in source control. (saveold)

RulesPacker/saveold/ApplicationInspector.CLI.runtimeconfig.json
```diff
@@ -0,0 +1,9 @@
+{
```
gfs, line 1: Same as previous, I don't think these files are supposed to be here.

UnitTest.Commands/source/empty.json
```diff
@@ -0,0 +1 @@
+2020-04-06 14:16:56 4 ERROR - AppInsLog - Command not supported for DLL calls
```
gfs, line 1: This file doesn't seem like it should be checked in

UnitTest.Commands/source/empty.cpp
```diff
@@ -0,0 +1 @@
+2020-04-06 14:15:16 1 ERROR - AppInsLog - Output file not specified for custom rules pack command
```
gfs, line 1: Same as others, why is this being checked in?

.gitignore
```diff
@@ -270,3 +270,4 @@ __pycache__/
 /Microsoft.DevSkim/linux/ApplicationInspector
```
gfs, line 270: Is this a file we have that we need to ignore?

NewDesignExample.cs
```diff
@@ -0,0 +1,59 @@
+﻿///////////////////////// COMMANDS CORE
```
gfs, line 1: Do you need to check this in? Particularly at the src root?

AppInspector.CLI/ResultsWriter.cs
guyacosta: Not supported currently as there are several dependencies on the local files for icons, js files etc.  requiring copying of the html directories to the target output path.  Future investigation to consolidate all files into a single file but not this PR.

NewDesignExample.cs
guyacosta: Unintentional thanks.  Has been removed in new commit to this branch.

UnitTest.Commands/source/empty.json
guyacosta: The unit tests need a source file that will result in no scan matches.  For the TagTest command two different files are needed as it will reject testing for delta's against the same one hence both an empty and blank named source set of files are checked in.  These two files are copied however.  A new commit has removed unneeded versions of these files which already exists in the desired locations under source\unzipped\simple.

UnitTest.Commands/source/empty.cpp
guyacosta: Removed unneeded versions of these files which already exists in the desired locations under source\unzipped\simple

RulesPacker/saveold/ApplicationInspector.CLI.runtimeconfig.json
guyacosta: Agreed.  Corrected unintended checkin.

AppInspector.CLI/CLICmdOptions.cs
guyacosta: If there were more than two possible argument values then yes but in this case all that is needed is a binary yes/no and a boolean is correct.

UnitTest.Commands/source/empty.json
gfs: You could also use `Path.GetTempFileName()` to get a blank file. [Path.GetTempFileName](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.gettempfilename?view=netframework-4.8)

If indeed these files are intended to have no matches, having any content in them seems non optimal.

AppInspector.CLI/Program.cs
guyacosta: Close but for the OutputFormat and OutputFilePath which are only applicable for CLI operations not for NuGet callers where file output is not supported in lieu of returning objects that they can serialize to file as needed.  In addition, the CLI version is not inherited from the options set for Commands module since that would require liquidhtml dependency where it is not used/needed so that the right property markers could be added required for the CommandLineParser used only in the CLI.  The solution was to transfer only the required arguments to the CommandOptions used in the Commands module allowing the CLI to verify the output arguments separately and manage file output.

AppInspector.CLI/Program.cs
guyacosta: Agreed thanks.  Added a ToLower() as well in case argument was mixed or upper case.

AppInspector.CLI/CLICmdOptions.cs
```diff
@@ -2,46 +2,42 @@
...
-        [Option('e', "text-format", Required = false, HelpText = "Match text format specifiers", Default = "Tag:%T,Rule:%N,Ruleid:%R,Confidence:%X,File:%F,Sourcetype:%t,Line:%L,Sample:%m")]
-        public string TextOutputFormat { get; set; }
-
         [Option('r', "custom-rules-path", Required = false, HelpText = "Custom rules file or directory path")]
         public string CustomRulesPath { get; set; }
 
-        [Option('t', "tag-output-only", Required = false, HelpText = "Output only identified tags", Default = false)]
-        public bool SimpleTagsOnly { get; set; }
+        [Option('h', "match-depth", Required = false, HelpText = "First match or best match based on confidence level (first|best)", Default = "best")]
+        public string MatchDepth { get; set; }
```
gfs, line 40: The above comment should be marking this line.  You have not nullable annotated these classes.

AppInspector.CLI/CLICmdOptions.cs
gfs: Sorry, see other comment. This actually refers to the line below, which is a `string` but you're not setting a default value so it is actually a `string?`.  Since this is a big rewrite it seems like a good opportunity to do nullable annotations.

AppInspector.CLI/Program.cs
guyacosta: Commit added to resolve.

AppInspector.CLI/Program.cs
guyacosta: interesting let's do it..

AppInspector.CLI/Program.cs
guyacosta: Not seeing this was likely removed by the code clean up commit added after your review.

AppInspector.CLI/Program.cs
guyacosta: Already added to commit with code clean up.

AppInspector.CLI/ResultsWriter.cs
guyacosta: agreed was as a left over from moving the code around.

AppInspector.CLI/Program.cs
```diff
@@ -25,103 +26,364 @@ static int Main(string[] args)
...
+            //final exit msg to review log
+            if (finalResult == (int)Utils.ExitCode.CriticalError)
             {
                 if (!String.IsNullOrEmpty(Utils.LogFilePath))
-                    WriteOnce.Info(ErrMsg.FormatString(ErrMsg.ID.RUNTIME_ERROR_UNNAMED, Utils.LogFilePath), true, WriteOnce.ConsoleVerbosity.Low, false);
+                    WriteOnce.Info(MsgHelp.FormatString(MsgHelp.ID.RUNTIME_ERROR_UNNAMED, Utils.LogFilePath), true, WriteOnce.ConsoleVerbosity.Low, false);
                 else
-                    WriteOnce.Info(ErrMsg.GetString(ErrMsg.ID.RUNTIME_ERROR_PRELOG), true, WriteOnce.ConsoleVerbosity.Medium, false);
+                    WriteOnce.Info(MsgHelp.GetString(MsgHelp.ID.RUNTIME_ERROR_PRELOG), true, WriteOnce.ConsoleVerbosity.Medium, false);
             }
```
gfs, line 71: Extra whitespace here

AppInspector.CLI/Program.cs
```diff
@@ -25,103 +26,364 @@ static int Main(string[] args)
...
 
 
             return finalResult;
         }
```
gfs, line 75: Extra whitespace here

AppInspector.CLI/Program.cs
gfs: I see two blank lines on 252 and 253 and 2 blank lines on 249 and 250. That's in addition to a blank line on 247 which is inconsistent with the style on 228/229.  I think you may need to double check what rules your code cleanup is applying because I don't think it is catching all the extra whitespace.

AppInspector.CLI/Writers/DummyWriters.cs
guyacosta: The json writers in most cases are the same but the text writers are distinct by type with different checks and formatting in some cases like the analyze writer including its html writer.  While I could consolidate a few of the json writers, it has some consistency to hold one per format and command type for identifying where that work is done and future distinctions for type.  I'm open to considering reducing where possible -let me chew on.

gfs: You could remove the boilerplate code here (and keep the classes if you want them) by changing the functions in the base CommandResultsWriter class to just throw NotImplemented (which I assume is the base case) and then you don't need all these overrides - alternately have a "DummyCommandResultsWriter" that has the notimplementedexceptions and then make subclasses of that.

AppInspector.CLI/Writers/VerifyRulesJsonWriter.cs
gfs: It seems like you could have fewer of these CommandResultsWriter classes since so few of them differ in behavior.  Or alternately use subclasses and inheritance to only implement these behaviors once.  Its a lot of reused code.

AppInspector.CLI/Writers/VerifyRulesJsonWriter.cs
guyacosta: For most json writers other than analyze yes.  I'll take care of to reduce.

AppInspector.CLI/Writers/VerifyRulesJsonWriter.cs
guyacosta: good catch.

AppInspector.CLI/Writers/VerifyRulesJsonWriter.cs
guyacosta: Yeah, that would reduce classes -see prior comments also for background but I'll do that.

AppInspector.CLI/Program.cs
guyacosta: found and fixed

AppInspector.CLI/Writers/VerifyRulesJsonWriter.cs
guyacosta: fixed

AppInspector.CLI/Writers/VerifyRulesJsonWriter.cs
guyacosta: fixed

AppInspector.CLI/ResultsWriter.cs
guyacosta: fixed

AppInspector.CLI/ResultsWriter.cs
guyacosta: fixed

AppInspector.CLI/CLICmdOptions.cs
guyacosta: Ah -fixed.  Yes on nullables but separating refactoring first as per priorities discussion but it's next

guyacosta: For CLI the defaults specified in the properties do the work but in cases where the property is overridden from the base the CommandLineParser doesn't set them right so I do in the derived class constructors.  This is fixed for all missing in this class.

RulesEngine/RuleProcessor.cs
guyacosta: It already passed see line 173 for being added as a tag the additional checks only apply to filtering for unique first/best settings.

AppInspector.CLI/CLICmdOptions.cs
guyacosta: have a separate issue for that effort which will be medium complex and want to separate it from this design refactor.

AppInspector.CLI/Writers/WriterFactory.cs
guyacosta: Agreed.  Fixed.

AppInspector.CLI/CLICmdOptions.cs
guyacosta: sweet trick will get on it

AppInspector.CLI/CLICmdOptions.cs
guyacosta: Because found that CommandLineParser will not set them if a new override is provided with a different default than the base.  Had to counter by putting in manually.

RulesEngine/RuleProcessor.cs
guyacosta: reads better -thx

PR #199, opened 2020-04-27
Log config flex

guyacosta:


guyacosta:
Provides the flexibility to default to a new config file or reuse an existing one as well as simply take a caller defined log complete with configuration.

PR #204, opened 2020-05-07
Performance Improvements - Parallelize Analysis, Extraction, Compile Regexes

### LINKED ISSUE #205: Compile Regexes and Cache them for performance
Each call to the regex constructor is expensive as the constructor has to parse the string into its logical regex representation.  If you want maximum performance out of each call to the regex you also want to use `RegexOptions.Compiled` which further increases that cost.

Creating a new regex object for each call to match pattern like in `TextContainer.cs` is very expensive. https://github.com/microsoft/ApplicationInspector/blob/71df07821b3ce7c9b4c5fe77ac4c7e56c68f4c78/RulesEngine/TextContainer.cs#L161 

Instead the `SearchPattern` when created should have a compiled regex inside of it.

.NET 5 Has extra improvements for compiled regexes (and normal regexes) as well. https://devblogs.microsoft.com/dotnet/regex-performance-improvements-in-net-5/

gfs:
Rewrite AnalyzeCommand to process files in parallel.
Made the stats in Metadata threadsafe.
Compile regexes for performance. Fix #205 
Update multiextractor to work in parallel when possible. *
Update multiextractor to catch exceptions. *

* I also contributed better implementations of these upstream to multiextractor in OSSGadget which we can incorporate once that is published as a nuget. microsoft/OSSGadget#47


gfs:
It looks like two tests are failing. `ExpectedTagCountDupsAllowed` and `ExpectedTagCountNoDupsAllowed`.

I pushed a new version of `ExpectedTagCountDupsAllowed` that checks both single and multi-threaded versions and it also fails on single threaded.

I looked into the code for what I might have changed related to dups/no dups and saw that it gets passed to metadata helper, but meta data helper just ignores it: https://github.com/microsoft/ApplicationInspector/blob/6375d1d73c594645e7eb863e03b6c4bed5164f1a/AppInspector/MetaDataHelper.cs#L22.

It also gets passed into RuleProcessor, but it doesn't look like I had to modify that file. https://github.com/microsoft/ApplicationInspector/blob/6375d1d73c594645e7eb863e03b6c4bed5164f1a/RulesEngine/RuleProcessor.cs#L44.

Any ideas?

gfs:
@guyacosta trying to understand how the deduping process works and it looks like it flows through `UniqueTagsCheck` https://github.com/microsoft/ApplicationInspector/blob/6375d1d73c594645e7eb863e03b6c4bed5164f1a/RulesEngine/RuleProcessor.cs#L318

I'm trying to understand this function but it looks like it bails out as invalid if ANY of the tags included are non-unique.  Wouldn't that then cause you to drop the other tags which may be unique since you can pass this an array of tags.

For example, if I pass two unique and one non unique tag this will return `false` here https://github.com/microsoft/ApplicationInspector/blob/6375d1d73c594645e7eb863e03b6c4bed5164f1a/RulesEngine/RuleProcessor.cs#L156

Causing `passedConditions` to be false.  But are the other two valid tags attached to the result.  Am I missing something?

AppInspector/Commands/AnalyzeCommand.cs
```diff
@@ -314,8 +314,8 @@ public AnalyzeResult GetResult()
                     catch (Exception e)
                     {
                         WriteOnce.SafeLog(e.Message + "\n" + e.StackTrace, LogLevel.Error);//log details
-                        Exception f = new Exception(MsgHelp.FormatString(MsgHelp.ID.ANALYZE_FILE_TYPE_OPEN, filename));//report friendly version
-                        throw f;
+                        //Exception f = new Exception(MsgHelp.FormatString(MsgHelp.ID.ANALYZE_FILE_TYPE_OPEN, filename));//report friendly version
+                        //throw f;
                     }
```
guyacosta, line 319: So this is allowing all files to be processed that will and just logging the failures rather than halt for a single failure...yeah, that's good and is actually an issue/goal.  We can take the commented code out for hygiene.

AppInspector/Commands/AnalyzeCommand.cs
```diff
@@ -641,27 +654,51 @@ private void UnZipAndProcess(string filename, ArchiveFileType archiveFileType, b
 
             try
             {
-                IEnumerable<FileEntry> files = Extractor.ExtractFile(filename);
+                IEnumerable<FileEntry> files = Extractor.ExtractFile(filename).Where(x => x != null);
```
guyacosta, line 657: What's the benefit on like 657?  I don't think you would get back a FileEntry object in the files that was null unless it throw an exception anyway.

gfs:
I'm not sure exactly how, but I was getting null FileEntries back in some cases. There shouldn't be any benefit except we can't 100% trust Extractor to not return nulls.

AppInspector/Commands/AnalyzeCommand.cs
gfs: I'm not sure exactly how, but I was getting null FileEntries back in some cases. There shouldn't be any benefit except we can't 100% trust Extractor to not return nulls.

AppInspector/Commands/AnalyzeCommand.cs
```diff
@@ -641,27 +654,51 @@ private void UnZipAndProcess(string filename, ArchiveFileType archiveFileType, b
...
+                                byte[] streamByteArray = file.Content.ToArray();
+                                ProcessInMemory(file.FullPath, Encoding.UTF8.GetString(streamByteArray, 0, streamByteArray.Length), languageInfo);
+                            }
+                        }
+                        catch (Exception)
                         {
-                            byte[] streamByteArray = file.Content.ToArray();
-                            ProcessInMemory(file.FullPath, Encoding.UTF8.GetString(streamByteArray, 0, streamByteArray.Length), languageInfo);
+                            Console.WriteLine($"Failed to parse {file.FullPath}");
                         }
```
guyacosta, line 676: Change msg to Failed to decompress file {file.FullPath} not parse for user clarity and change Console.WriteLine WriteOnce.SafeLog.  I think since we're handling parrellel processing now the final output should include a msg to check the logs for any notices and/or track if there were non-fatal exceptions written to the log and let them know.

gfs: Fixed in 560b8bf.

gfs:
Responded to all comments.  Added regex compilation functionality.  

@guyacosta Can you take a look at the failing tests? I don't understand the failing test cases well enough to fix them.  Could it be related to #207?

guyacosta:
Reviewing latest changes.

guyacosta:
@gfs this is still coming up with unit test failures 21 now so please review and correct so that the PR is ready for final review and merge.  If there are 1-2 that you need help with let me know.

gfs:
Missed that Will do. I'll ping you when I've made some progress.

-------- Original Message --------

gfs:
@guyacosta I spent some time debugging the tests and here's my findings so far.

1. Good news: All of the library tests pass. All the failing tests are in the CLI.
2. Weird news: If I run some of the failing tests individually they don't fail.  They only fail when run in sequence with the other tests.  I'm not sure what to make of that.  Some of them also are intermittent in pass/fail.  However,
3. The CLI based tests are actually impossible to debug because they process call a build of the application rather than calling the APIs. Have you considered calling `main` with the command line args instead?

Is there some way to debug those tests I'm missing?  I'm kind of stuck here without failures in tests that are debuggable.

gfs:
The only test that I keeps failing for me when run individually is
`RulesNotPresentNoResults_Fail` and the exit code coming out is `-532462766`

guyacosta:
Yeah basically you take the arguments and feed them into the cly through the debugger
________________________________

gfs:
Should I just rewrite the tests to do that so they are inherently debuggable?

Sent from ProtonMail mobile

-------- Original Message --------

guyacosta:
still getting 21 not passing.  let me try to work through them so we can get this merged.

guyacosta:
done...

PR #208, opened 2020-05-18
Fix #207

### LINKED ISSUE #207: Requesting Unique Tags may result in an incomplete list of tags
**Describe the bug**
If you have Rule One with Tags `A` and `C`  and Rule Two with Tags `A` and `B`. 
If `unique tags only` is on and
* Rule One matches first: Two will not match and results will not be tagged with `B`.
* Rule Two matches first, One will not match and the results will not be tagged with `C`.

In #204 we parallelize the analysis so the order rules are applied becomes non-deterministic and you will get either behavior 1 or 2 above at random.

**To Reproduce**
Pass appropriate values to `UniqueTagsCheck` if any of the tags passed in are included in the set of known tags all tags will be ignored. https://github.com/microsoft/ApplicationInspector/blob/6375d1d73c594645e7eb863e03b6c4bed5164f1a/RulesEngine/RuleProcessor.cs#L324
**Expected behavior**
Requesting unique tags should return all matched tags, deduplicated.

**Proposed Fix**
We should check if there are any new tags and grab them, rather than checking for duplicate tags and discarding everything.

guyacosta:


RulesEngine/RuleProcessor.cs
```diff
@@ -310,45 +310,36 @@ private IEnumerable<Rule> GetRulesForLanguages(string[] languages)
...
-                    if (_logger != null && !approved)
-                    {
-                        _logger.Debug(string.Format("Duplicate tag {0} not added", tag));
-                    }
-
-                    break;
+                    }                   
                 }
-                else
+                else if (_logger != null && !approved)
```
gfs, line 340: You don't know if its a duplicate until you've checked everything and determined if its approved or not. Which might occur inside of one of if branches this was attached to.

```suggestion
                
                if (_logger != null && !approved)
```

PR #217, opened 2020-05-22
Remove unnecessary call to Match.count (small perf improvement).

scovetta:
This PR includes a small optimization - If you use Match.count, it immediately runs the regular expression and collects all matches. It isn't necessary in this case, since there are times when we stop processing (`_stopAfterFirstMatch` or too many pattern matches). Just using `foreach (Match m in ...)` makes this evaluation lazy, and in my testing, there was a 8-10% performance increase.

I also renamed the `bool MatchPattern(...)` overload to `IsMatchPattern` because the other overloads returned a List.


gfs:
LGTM.

gfs:
I have some improvements to this coming (including making the same changes in the other TextContainer.cs). Don't merge yet.

gfs:
All done.

PR #222, opened 2020-05-28
Fix Multiextractor include

gfs:
Currently including the commands nupkg gives an error from missing multiextractor.  This should fix it.

Also updates to the latest multiextractor from OSS Gadget.


gfs:
Fix #223

PR #225, opened 2020-05-29
Create SUPPORT.md

scovetta:


jhornick-ms:
We are piloting a new support statement, to make it clear to repo users where/how to ask questions about the usage of the software in this repo.  Our goal is to eventually have a statement like this in most Microsoft repos.  Do you have any feedback about the support statement/document?  Please share it here.

PR #230, opened 2020-06-16
Adds sdl only pipeline with codeinspector task+

guyacosta:


guyacosta:
Going to merge it due to branch rename efforts so build works but you can review later.

PR #232, opened 2020-06-16
Update references from master to main.

scovetta:
Found a few references to 'master', updated these both to 'main'. I'm not *sure* that the change is version.json is correct, @daalcant  Could you take a look?


guyacosta:
Good catch -thx.

PR #239, opened 2020-06-22
Update Pipelines with Templates

daalcant:


Pipelines/templates/dotnet-publish-nix-job.yml
```diff
@@ -0,0 +1,96 @@
...
+      archiveType: 'zip'
+      archiveFile: 'Archives/${{ parameters.projectName }}_Nix.zip'
+      replaceExistingArchive: true
+  - task: PublishBuildArtifacts@1
+    displayName: Pipeline Publish Archive
+    inputs:
+      PathtoPublish: 'Archives'
+      ArtifactName: '${{ parameters.artifactName }}'
+      publishLocation: 'Container'
+
```
gfs, line 93: Whitespace

Pipelines/templates/dotnet-publish-nix-job.yml
```diff
@@ -0,0 +1,96 @@
+parameters:
+# Job Name
+- name: jobName
+  type: string
+  default: 'publish_nix'
```
gfs, line 5: Either this should be named approrpiately (nix-mac-job) or it should only do linux with a separate job that does mac

Pipelines/templates/dotnet-publish-win-netcore-job.yml
```diff
@@ -0,0 +1,107 @@
...
+  - task: PublishBuildArtifacts@1
+    displayName: Pipeline Publish Archive
+    inputs:
+      PathtoPublish: 'Archives'
+      ArtifactName: '${{ parameters.artifactName }}'
+      publishLocation: 'Container'
+
+
+
+
```
gfs, line 107: Whitespace

Pipelines/templates/dotnet-test-job.yml
```diff
@@ -0,0 +1,44 @@
...
+  displayName: Dotnet Test
+  pool:
+    vmImage: 'windows-latest'
+  steps:
+  - task: UseDotNet@2
+    displayName: Install Dotnet SDK
+    inputs:
+      packageType: 'sdk'
+      version: ${{ parameters.dotnetVersion }}
+  - task: NuGetToolInstaller@1
```
gfs, line 30: why install nuget if we're using dotnet restore?

Pipelines/templates/dotnet-publish-win-netcore-job.yml
```diff
@@ -0,0 +1,107 @@
...
+      version: ${{ parameters.dotnetVersion }}
+  - task: NuGetToolInstaller@1
+    displayName: Install Nuget Tool
+    inputs:
+      versionSpec: ${{ parameters.nugetVersion }}
+  - bash: 'echo "##vso[task.setvariable variable=ReleaseVersion;]$RELEASE"'
+    env:
+      RELEASE: ${{ parameters.releaseVersion }}
+    displayName: Set Release Version
+  - ${{ parameters.preBuild }}
```
gfs, line 58: The prebuild runs here to set the version but it looks like you set the version already on line 54.

It looks like we don't have to do any manual version trickery for the nuget ones.  I'm interested in why each works each way.

Pipelines/templates/dotnet-publish-nix-job.yml
```diff
@@ -0,0 +1,96 @@
...
+      version: ${{ parameters.dotnetVersion }}
+  - task: NuGetToolInstaller@1
+    displayName: Install Nuget Tool
+    inputs:
+      versionSpec: ${{ parameters.nugetVersion }}        
+  - bash: 'echo "##vso[task.setvariable variable=ReleaseVersion;]$RELEASE"'
+    env:
+      RELEASE: ${{ parameters.releaseVersion }}
+    displayName: Set Release Version
+  - ${{ parameters.preBuild }}
```
gfs, line 58: you set the version before you run the prebuild task that sets the version here.  Am I missing something?

gfs:
Looks super promising!

Pipelines/templates/dotnet-publish-nix-job.yml
daalcant: Removed

Pipelines/templates/dotnet-publish-nix-job.yml
daalcant: Using nix as a loose term here for Unix/Linux-style systems. We could create a separate job for Mac, but this is a bit faster. Easy enough to split in the future if we see a need for it.

daalcant: Not opposed to renaming the default jobName though.

Pipelines/templates/dotnet-publish-win-netcore-job.yml
daalcant: Removed

Pipelines/templates/dotnet-test-job.yml
daalcant: Pretty sure 'dotnet restore' still leans on NuGet. This is for some past bugs we encountered where random packages weren't restoring correctly, and updating the NuGet tool before running 'dotnet restore' resolved the problem.

Pipelines/templates/dotnet-publish-win-netcore-job.yml
daalcant: Doing this to provide multiple ways of setting the release version. Not immediately useful, but may need it in the future.

Pipelines/templates/dotnet-publish-nix-job.yml
gfs: I think just rename the default job. The windows one is called "windows_netcore" or whatever

Pipelines/templates/dotnet-publish-nix-job.yml
daalcant: Doing this to provide multiple ways of setting the release version. Not immediately useful, but may need it in the future.

Pipelines/templates/dotnet-publish-win-netcore-job.yml
gfs: I think we only need one for now... I would choose the one that is actually being used and integrate that into the template.

Pipelines/templates/dotnet-publish-nix-job.yml
daalcant: Did this, as well as changed the file name for consistency.

Pipelines/templates/dotnet-publish-win-netcore-job.yml
daalcant: I'll remove the release version being set within the template then.

daalcant: Updated

Issue #250, opened 2020-07-27
UniqueDependencies list found in match details


guyacosta:
Detected bug with UniqueDependencies list being empty for some cases and produces  "uniqueDependencies": {}, in json for instance.  When found all instances are included in the list of detailed matches.   

Desired: report tag "Dependency.SourceInclude" in uniqueTags result but does not include those matches in match details list i.e. we still don't want to report each one and where to avoid creating noise; same as original design goal.

guyacosta:
Fixed in PR

guyacosta:
Putting in a separate PR to come

guyacosta:
#246

Issue #235, opened 2020-06-16
Add Rust specific rules patterns


guyacosta:
Related to #234 

This is to prioritize and add Rust language specific rules similar to those used now for C++ support.

PR #246, opened 2020-07-26
Guyacosta/sorteddictionaries

guyacosta:
Primary motivation is to fix #245 which describes result consistency issues, html report errors and unnecessary data fields as a left over from internal concurrency work.  This effort retains that significant improvement while fixing a few incidental issues including the additional ones below in comments.

PR #251, opened 2020-07-27
Rust support

guyacosta:
Adds Rust language support #235 
Fixes an issue #250 with UniqueDependencies list from removed logic to avoid adding records to detailed match list when found.

### DIFF
--- a/AppInspector.CLI/Writers/AnalyzeHtmlWriter.cs
+++ b/AppInspector.CLI/Writers/AnalyzeHtmlWriter.cs
@@ -184,7 +184,7 @@ namespace Microsoft.ApplicationInspector.CLI
                 {
                     foreach (TagSearchPattern pattern in tagGroup.Patterns)
                     {
-                        pattern.Detected = _appMetaData.UniqueTags.Any(v => v.Key.Contains(pattern.SearchPattern));
+                        pattern.Detected = _appMetaData.UniqueTags.Any(v => v.Contains(pattern.SearchPattern));
                         //create dynamic "category" groups of tags with pattern relationship established from TagReportGroups.json
                         //that can be used to populate reports with various attributes for each tag detected
                         if (pattern.Detected)
@@ -422,12 +422,10 @@ namespace Microsoft.ApplicationInspector.CLI
         /// <returns></returns>
         private List<TagInfo> GetTagInfoListByName()
         {
-            List<string> orderedTags = _appMetaData.UniqueTags.Keys.ToList<string>();
-            orderedTags.Sort();
             HashSet<string> dupCheck = new HashSet<string>();
             List<TagInfo> result = new List<TagInfo>();
 
-            foreach (string tag in orderedTags)
+            foreach (string tag in _appMetaData.UniqueTags)
             {
                 foreach (var match in _appMetaData.Matches)
                 {
@@ -465,9 +463,8 @@ namespace Microsoft.ApplicationInspector.CLI
             List<TagInfo> result = new List<TagInfo>();
             HashSet<string> dupCheck = new HashSet<string>();
             RulesEngine.Confidence[] confidences = { Confidence.High, Confidence.Medium, Confidence.Low };
-
-            
-            foreach (string tag in _appMetaData.UniqueTags.Keys)
+       
+            foreach (string tag in _appMetaData.UniqueTags)
             {
                 var searchPattern = new Regex(tag, RegexOptions.IgnoreCase);
                 foreach (Confidence test in confidences)
@@ -509,8 +506,7 @@ namespace Microsoft.ApplicationInspector.CLI
             HashSet<string> dupCheck = new HashSet<string>();
             RulesEngine.Severity[] severities = { Severity.Critical, Severity.Important, Severity.Moderate, Severity.BestPractice, Severity.ManualReview };
 
-
-            foreach (string tag in _appMetaData.UniqueTags.Keys)
+            foreach (string tag in _appMetaData.UniqueTags)
             {
                 // TODO: How frequently are these generated? Cache?
                 var searchPattern = new Regex(tag, RegexOptions.IgnoreCase);
--- a/AppInspector.CLI/Writers/AnalyzeHtmlWriter.cs
+++ b/AppInspector.CLI/Writers/AnalyzeHtmlWriter.cs
@@ -89,7 +89,7 @@ namespace Microsoft.ApplicationInspector.CLI
             hashData["ostargets"] = _appMetaData.OSTargets;
             hashData["outputs"] = _appMetaData.Outputs;
             hashData["filetypes"] = _appMetaData.FileExtensions;
-            hashData["tagcounters"] = ConvertTagCounters(_appMetaData.TagCounters.Values);
+            hashData["tagcounters"] = ConvertTagCounters(_appMetaData.TagCounters);
 
             //final render and close
             var htmlResult = htmlTemplate.Render(hashData);
--- a/AppInspector.CLI/Writers/AnalyzeHtmlWriter.cs
+++ b/AppInspector.CLI/Writers/AnalyzeHtmlWriter.cs
@@ -11,6 +11,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Text;
 using System.Text.RegularExpressions;
 
 namespace Microsoft.ApplicationInspector.CLI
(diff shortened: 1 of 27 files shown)


AppInspector.CLI/Writers/AnalyzeHtmlWriter.cs
```diff
@@ -107,6 +114,39 @@ private void WriteJsonResult()
...
+        private string MergeResourceFiles(string inputPath)
+        {
+            StringBuilder stringBuilder = new StringBuilder();
+
+            if (Directory.Exists(inputPath))
+            {
+                try
+                {
+                    IEnumerable<string> srcfileList = Directory.EnumerateFiles(inputPath, "*.*", SearchOption.AllDirectories);
+                    if (srcfileList.Count() == 0)
```
gfs, line 128: Use .Any

AppInspector.CLI/Writers/AnalyzeJsonWriter.cs
```diff
@@ -40,8 +40,8 @@ public override void WriteResults(Result result, CLICommandOptions commandOption
 
             if (cLIAnalyzeCmdOptions.SimpleTagsOnly)
             {
-                List<string> keys = new List<string>(analyzeResult.Metadata.UniqueTags.Keys);
-                keys.Sort();
+                List<string> keys = new List<string>(analyzeResult.Metadata.UniqueTags);
+                //keys.Sort();
```
gfs, line 44: Generally try not to check in commented code

AppInspector/Commands/AnalyzeCommand.cs
```diff
@@ -75,7 +75,7 @@ public class AnalyzeCommand
         private DateTime _lastUpdated;
 
         //save all unique dependencies even if Dependency tag pattern is not-unique
-        private Regex tagPatternRegex = new Regex("Dependency.SourceInclude", RegexOptions.IgnoreCase);
+        private Regex tagDepPatternRegex = new Regex("Dependency.SourceInclude", RegexOptions.IgnoreCase);
```
gfs, line 78: If this runs a lot you should pass `RegexOptions.Compiled`

AppInspector/MetaData.cs
```diff
@@ -206,26 +199,18 @@ public MetaData(string applicationName, string sourcePath)
...
-                ["strGrpTargets"] = new ConcurrentDictionary<string,byte>(),
-                ["strGrpOSTargets"] = new ConcurrentDictionary<string,byte>(),
-                ["strGrpFileExtensions"] = new ConcurrentDictionary<string,byte>(),
-                ["strGrpFileNames"] = new ConcurrentDictionary<string,byte>(),
-                ["strGrpCPUTargets"] = new ConcurrentDictionary<string,byte>(),
-                ["strGrpCloudTargets"] = new ConcurrentDictionary<string,byte>(),
-                ["strGrpUniqueDependencies"] = new ConcurrentDictionary<string,byte>()
-            };
-
+            PackageTypes = new List<string>();
```
gfs, line 202: Why not declare these inline with the declaration instead of in the constructor?  Would keep the constructor cleaner.

AppInspector/MetaDataHelper.cs
```diff
@@ -11,11 +13,22 @@ namespace Microsoft.ApplicationInspector.Commands
...
+        public ConcurrentDictionary<string, byte> FileNames { get; set; }
+        public ConcurrentDictionary<string, byte> UniqueTags { get; set; }
+        public ConcurrentDictionary<string, byte> UniqueDependencies { get; set; }
+        public ConcurrentDictionary<string, byte> Outputs { get; set; }
+        public ConcurrentDictionary<string, byte> Targets { get; set; }
+        public ConcurrentDictionary<string, byte> FileExtensions { get; set; }
+        public ConcurrentDictionary<string, byte> CPUTargets { get; set; }
+        public ConcurrentDictionary<string, byte> CloudTargets { get; set; }        
+        public ConcurrentDictionary<string, byte> OSTargets { get; set; }
+        ConcurrentDictionary<string,MetricTagCounter> TagCounters { get; set; }
```
gfs, line 31: why is this one not public but the others are?

Are these supposed to be public in the first place?  Shouldn't they be internal since this is designed to hide complexity from users?

AppInspector/MetaDataHelper.cs
```diff
@@ -24,13 +37,19 @@ public MetaDataHelper(string sourcePath, bool uniqueMatchesOnly)
...
             Metadata = new MetaData(GetDefaultProjectName(sourcePath), sourcePath);
 
-            _propertyTagSearchPatterns = new Dictionary<string, Regex>()
-            {
-                { "strGrpOSTargets", new Regex(".OS.Targets", RegexOptions.Compiled | RegexOptions.IgnoreCase) },
-                { "strGrpCloudTargets", new Regex(".OS.Targets", RegexOptions.Compiled | RegexOptions.IgnoreCase) },
-                { "strGrpOutputs", new Regex(".OS.Targets", RegexOptions.Compiled | RegexOptions.IgnoreCase) },
-                { "strGrpCPUTargets", new Regex(".OS.Targets", RegexOptions.Compiled | RegexOptions.IgnoreCase) }
-            };
+            PackageTypes = new ConcurrentDictionary<string, byte>();
```
gfs, line 40: Declare inline with variables.

AppInspector/MetaDataHelper.cs
```diff
@@ -132,6 +179,8 @@ public void AddLanguage(string language)
             Metadata.Languages.AddOrUpdate(language, 1, (language, count) => count + 1);
         }
 
+        #region helpers
```
gfs, line 182: region appears to start and end at different scopes

AppInspector/MetaDataHelper.cs
```diff
@@ -112,17 +133,43 @@ public void AddMatchRecord(MatchRecord matchRecord)
...
             {
                 Metadata.IncrementTotalMatchesCount(-1);//reduce e.g. tag counters not included as detailed match
             }
         }
 
+
+        /// <summary>
+        /// Transfer concurrent data from scan to analyze result with sorted, simplier types for callers
+        /// </summary>
+        public void PrepareReport()
```
gfs, line 154: Instead of Metadata being a static class shouldn't this return the serializable `Metadata` object?

AppInspector.CLI/Writers/AnalyzeJsonWriter.cs
guyacosta: Agreed.  Sort no longer needed and should be completed removed -will fix.

AppInspector/MetaDataHelper.cs
guyacosta: Some are visible from AnalyzeCommand, others aren't and are processed when a match is added.  I'll recheck each.

gfs: Analyze command is in the same namespace so it shouldn't need to be public to access it.

Public means people writing code using the app inspector library can access these properties (and they should be documented and supported)

Do you intend to publicly expose the dictionaries? I thought not.

guyacosta:
Will resubmit based on main to more easily distinguish from as yet approved PR #246

AppInspector.CLI/Writers/AnalyzeHtmlWriter.cs
scovetta: (Advantage of .Any() is that it stops after the first match, so for this, it should be *much* faster.)

AppInspector.CLI/Writers/AnalyzeJsonWriter.cs
scovetta: Isn't UniqueTags a ConcurrentDictionary - e.g. you need .Keys in there somewhere. You should be able to simplify this with:

```
TagsFile tags = new TagsFile() {
  Tags = analyzeResult.Metadata.UniqueTags.Keys.ToArray()
};
```

AppInspector/Commands/AnalyzeCommand.cs
scovetta: You don't need to use a Regex here -- just do a string contains. If there's another reason for using the regex, just escape the period in the middle so it isn't interpreted.

AppInspector.CLI/Writers/AnalyzeJsonWriter.cs
guyacosta: This PR was built based on #246 which has modified the data structures.  Gabe's comments thus are representing both sets of changes.  I have closed PR 251 as requested and submitted a new one based on main to help distinguish the differences better.  Please see #246 first.  Thanks.

AppInspector/MetaDataHelper.cs
guyacosta: internal works...thanks

AppInspector/MetaDataHelper.cs
guyacosta: If I did I'd have to wrap calls for 3 of the data types for incrementing counts which seems needless.

gfs: I'm not sure I understand why that would involve more call wrapping.

Your analyzer function could in theory be called twice at once and using a static class to hold non-static data like this is just inviting weird behavior.

guyacosta: Metadata is not a static class.  If you see it add the code line here.  What I responded interpreted was return a Metadata object from PrepareReport but an instance exists before that and is needed.  Also the AnalyzeCommand should only be called once by a given caller by instance until it completes though the scan is threaded.

AppInspector.CLI/Writers/AnalyzeHtmlWriter.cs
guyacosta: Yeah good call.  The function where this will be used is temporarily removed but will be reintroduced as a separate PR focused on resource output consolidation and I'll use that approach for similar needs.

Issue #253, opened 2020-07-27
Review all rule patterns for use of case insensitivity correctness


guyacosta:
Review all rule patterns for use of case insensitivity correctness.  Some cases like specific API's may require sensitive comparison.

guyacosta:
- API specific patterns should be case insensitive and most already were but a few had the insensitive modifier needlessly
- General patterns that may occur across multiple variations of api's (unnamed) are used to "catch" those usages for which there are too many possibilities
- Case insensitive patterns add to search performance impacts thus balancing the specific vs general patterns is get the best speed while not missing valid feature identification opportunities improves the default behavior.

PR #256, opened 2020-08-03
#253 - Few case insenstitive modifiers were removed as unnecessary for…

guyacosta:
… specific API patterns.  Couple of pattern corrections made also.

### DIFF
--- a/AppInspector/rules/default/components/active_content.json
+++ b/AppInspector/rules/default/components/active_content.json
@@ -64,11 +64,19 @@
           "code",
           "comment"
         ],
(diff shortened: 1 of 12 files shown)


guyacosta:
- API specific patterns should be case insensitive
- General patterns that may occur across multiple variations of api's (unnamed) are used to "catch" those usages for which there are too many possibilities
- Case insensitive patterns add to search performance impacts thus balancing the specific vs general patterns is get the best speed while not missing valid feature identification opportunities improves the default behavior.

Issue #258, opened 2020-08-04
Add Android intents scanning


scovetta:
We should add detection for Android [Intents](https://developer.android.com/guide/components/intents-common) to Application Inspector.

gfs:
For iOS how this works might be interesting
https://github.com/delba/Permission

gfs:
For BSD you could look at pledges: https://man.openbsd.org/pledge.2

gfs:
UWP: https://docs.microsoft.com/en-us/windows/uwp/packaging/app-capability-declarations

guyacosta:
I'm thinking the IOS, BSD and UWP are all in a somewhat related but distinct category i.e. permissions request/grant rather than the general request for a custom operation to be invoked through a general process create call or Intents which indirectly achieves a similar result if a process exists to service the request.  I'll move the permissions related support request to another Issue.

scovetta:
I think we should add the specific intents instead of just usage of intents -- meaning things like ACTION_SET_ALARM. Might need to separate those out into the relevant category instead of bucketing them into a general intent tag.

guyacosta:
Note there a LOT of them. Will look them over this weekend to propose which specific intents to include and drop the list here for discussion.

guyacosta:
New support added for "Device.Permissions..." for Android, iOS, UWP for Java, Kotlin languages.

PR #260, opened 2020-08-06
#258 Android Intents rule support

guyacosta:
#258

### DIFF
--- a/AppInspector/rules/default/os/dynamic_execution.json
+++ b/AppInspector/rules/default/os/dynamic_execution.json
@@ -349,5 +349,86 @@
         "confidence": "high"
       }
     ]
+  },
+  {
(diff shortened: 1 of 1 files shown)


guyacosta:
Java, C++, C# support added and tested

guyacosta:
See #261 with split out enhancement request for device permissions detection and #262 for Kotlin language support request.  This PR covers existing supported languages and Intents only.

scovetta:
I think we should add the specific intents instead of just usage of intents -- meaning things like ACTION_SET_ALARM. Might need to separate those out into the relevant category instead of bucketing them into a general intent tag.

gfs:
Agree with Mike, all Android apps will use Intents in some way. What is interesting is what Intents they ask for.  I imagine many map directly to existing tags.

PR #271, opened 2020-08-15
One html file

### LINKED ISSUE #241: Locate html output to named/contained folder
The default html output file output.hmtl is located today in the application or runtime folder with dependencies located under the html subfolder(s).  To make the output more portable, write the output and dependencies to a single parent folder to make it easier to move it elsewhere or package it up for sending to someone.

### LINKED ISSUE #264: HTML Report Not Loading Correctly for NuGet CLI Install Option
### Steps:
` appinspector analyze -s /path/to/some/code`

### Expected:
Formatted Report

### Result:
Result is not formatted and is missing all images.

### Version:
`ApplicationInspector.CLI 1.2.49+c4843d26aa`

image

guyacosta:
Fixes #241 for local html report resources to be aggregated into the output file
Fixes #264 for application not working for html format when run from Nuget CLI install


guyacosta:
Allows user to specify an output file path now since dependencies are addressed and file can be located anywhere.

AppInspector.CLI/Writers/AnalyzeHtmlWriter.cs
```diff
@@ -51,9 +52,18 @@ public override void WriteResults(Result result, CLICommandOptions cLICommandOpt
 
         private void WriteHtmlResult()
         {
+            //Grab any local css and js files that are needed i.e. don't have hosted URL's or are proprietary
+            string allCSS = "<style>\n" + MergeResourceFiles(Utils.GetPath(Utils.AppPath.basePath) + "html\\resources\\css");
```
gfs, line 56: You should be using WebPack in the build or pre-build process to pack and minify these assets.

https://www.c-sharpcorner.com/article/getting-started-with-webpack/

AppInspector.CLI/Writers/AnalyzeHtmlWriter.cs
```diff
@@ -120,6 +130,37 @@ private void RegisterSafeType(Type type)
...
+                catch (Exception)
+                {
+                    throw new OpException(MsgHelp.FormatString(MsgHelp.ID.CMD_INVALID_FILE_OR_DIR, inputPath));
+                }
+            }
+
+            return stringBuilder.ToString();
+        }
+
+
```
gfs, line 163: Nit: whitespace

AppInspector.CLI/html/index.html
```diff
@@ -3,17 +3,17 @@
 <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
-    <link rel="stylesheet" href="html/resources/css/bootstrap.min.css" />
-    <link rel="stylesheet" href="html/resources/css/all.css" />
-    <link rel="stylesheet" href="html/resources/css/c3.min.css" />
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/4.5.2/css/bootstrap.min.css" />
```
gfs, line 6: This means you can't view the report offline and it still has external dependencies.  See comment about WebPack for properly minifying assets including dependencies like these.

AppInspector.CLI/html/index.html
```diff
@@ -51,16 +51,13 @@
...
-    <script src="html/resources/js/popper.min.js"></script>
-    <script src="html/resources/js/bootstrap.min.js"></script>
-    <script src="html/resources/js/ace.js"></script>
-    <script src="html/resources/js/ext-settings_menu.js"></script>
-    <script src="html/resources/js/ext-modelist.js"></script>
-    <script src="html/resources/js/ext-beautify.js"></script>
-    <script src="html/resources/js/d3.min.js"></script>
-    <script src="html/resources/js/c3.min.js"></script>
-    <script type="text/javascript" src="html/resources/js/appinspector.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.5.1/jquery.min.js" integrity="sha512-bLT0Qm9VnAYZDflyKcBaQ2gg0hSYNQrJ8RilYldYQ1FxQYoCLtUjuuRuZo+fjqhx/qtq/1itJ0C2ejDxltZVFg==" crossorigin="anonymous"></script>
```
gfs, line 54: See other comments about webpack.

guyacosta:
Cool I'll take a look at that option and get back to you.  Hey you should have dropped that in one the related issues for me earlier to save me time haha but thanks and sounds like a better approach!

gfs:
Sorry I didn't think to mention it earlier :P.

My recommendation is as follows.  This eliminates all dependency on external files being present on the disc, everything will be in the binary.

1. Use webpack either pre-build or as part of the pre-build build tasks to compile all your js and css assets.  It sounds like your html is also one file too, which is great.
2. Embed those into your binary.  Set the build task to EmbeddedResource in visual studio.
3. Have a simple test to ensure that the embeds are there.
4. At runtime access the embedded files like here in attack surface analyzer https://github.com/microsoft/AttackSurfaceAnalyzer/blob/5932999bf4b9cc4fd521aee1f84c0dcbc71de9a4/Lib/Objects/RuleFile.cs#L78-L108
5. Write the embedded files to disc where the user has selected.

gfs:
If you're not already, you should also embed your default rules, so the binary will be completely independent of any runtime data needed to operate.

guyacosta:
While the webpack does help consolidate js and cs files and removes the internet dependency for the files and attendant download time, the goal is to output a single file, not a reduced set of files.  Therefore, the application still needs to add the content into the index.html file to make it truly transportable without any additional local files that have to be included to zip up etc. which is my goal.  I see therefore the logic needed to achieve that merge as still necessary even with a proposed WebPack addition, with the change only improving the download performance which is negligible and do not have a goal for offline use -typical policy is most people have an active connection.  At the same time, the change represents a chunk of work and added depending on NPM, config, build and testing.  Looking at effort to value I'm thinking this is a refactor option in the near future but not necessary for the present and I need to get the CLI app working again asap for those that are unable to use it at present.  I ran this staged approach past management and have agreement to add this as a suggested engineering enhancement.  Thanks for the tip!

PR #281, opened 2020-09-04
Oat scan engine

guyacosta:
Final submission is pending investigation of these issues in the new Microsoft.CST.OAT dependency:
1. Expected rule captures are missing.  Compare old vs new attached outputs of example repo scan [azure-key-vault-to-kubernetes](https://github.com/SparebankenVest/azure-key-vault-to-kubernetes) using Release vs this branch and see lines 83-85 in RulesEngine\RulesProcessor.cs in new branch to uncomment and verify missing valid matches of expected tags->rules.
[AKV2KubOAT.json.txt](https://github.com/microsoft/ApplicationInspector/files/5173275/AKV2KubOAT.json.txt)
[AKV2KubB4OAT.json.txt](https://github.com/microsoft/ApplicationInspector/files/5173276/AKV2KubB4OAT.json.txt)
  a. Related: Rule AI007700 for CryptoCurrency will detect values in the first pattern set, but not the latter ones i.e. 'zcash' and 'monero' are not found while 'cryptocurrency' is found
  b. Related: dependency capture results are also impacted using new engine requiring investigation (see attached files also)
2. Lines 99-105 in RulesEngine\RulesProcessor.cs need a review for whether indicated logic should be included (not found
in DevSkim OAT version) to ensure non-Within captures are processed with Within captures exists.  If valid, needs correction in DevSkim also.
3. Proposed custom operation OATRegexWithIndexOperation to facilitate getting clause data index doesn't fire break point in delegate but this is more FYI as alternate method is working and clearer perhaps

General
98% of changes are in RulesEngine project.  In addition to the integration, nullable support was added to the RulesEngine project to be completed across the solution soon.  CLSCompliant (true) was removed from RulesProcessor.cs until it can be applied across the entire solution without warnings.


RulesEngine/RuleProcessor.cs
```diff
@@ -17,242 +16,235 @@ namespace Microsoft.ApplicationInspector.RulesEngine
...
+                                                    if (!_stopAfterFirstMatch) //use first or best match options
+                                                    {
+                                                        if (BestMatch(newMatches, newMatch) && BestMatch(resultsList, newMatch))
+                                                        {
+                                                            newMatches.Add(newMatch); //replaces previous and now removed match
+                                                        }
+                                                    }
+                                                    else
+                                                    {
+                                                        newMatches.Add(newMatch); //first unique match and we're done
```
gfs, line 160: This doesn't look right to me.  This adds the full set of new matches the same as the below else statement does.

RulesEngine/OATScopedRegexOperation.cs
```diff
@@ -0,0 +1,109 @@
...
+                                    if (tc.ScopeMatch(src.Scopes, translatedBoundary))
+                                    {
+                                        boundaries.Add(translatedBoundary);
+                                    }
+                                }
+                            }
+                        }
+
+                        var result = src.Invert ? boundaries.Count == 0 : boundaries.Count > 0;
+                        return new OperationResult(result, result && src.Capture ? new TypedClauseCapture<List<Boundary>>(c, boundaries, state1) : null);
```
gfs, line 63: This returns inside the foreach loop so it will always only use the first pattern.

gfs (changes requested):
1. I commented and identified why it was stopping on first match, if you copy pasted that direct from DevSkim that does mean there is a bug in DevSkim too.
2. See https://github.com/microsoft/DevSkim/blob/32d1e8696c0473042af888aa66d775168b5d7ff6/DevSkim-DotNet/Microsoft.DevSkim/RuleProcessor.cs#L125 for how I did this with ScopedRegex and Within for DevSkim.  This gets the captures from the WithinClause if there was one (i.e. from the matches that matched the within condition) and other clause if not (indicating there were no conditions)
3. Can you explain the alternate method?  For the delegate not firing I'll need to see the full example and the rule you were passing to say why it might not be firing, but all operations (including default) are assigned and called using the same lookup table so there shouldn't be any issue on the OAT side.

RulesEngine/OATScopedRegexOperation.cs
gfs: This was a bug in DevSkim as well that I've now fixed.

gfs (changes requested):

RulesEngine/OATScopedRegexOperation.cs
guyacosta: cool I'll grab from there

guyacosta:
Update.  The suggested custom operation delegate provided [here ](https://github.com/microsoft/OAT/issues/42)on the OAT issue site does not fire for reasons unknown but even copying the logic over to the ScopedRegexOperationDelegate method used in DevSkim which does get the event call fails to find any matches and I need more time to find out why around weekend events.  My alternative approach was to compare the Clause pattern in the boundary with those in the Rule search patterns to find a match then use the Confidence data and regex pattern from it to get the desired metadata etc.  I did have a small error in the logic and need more time to finish one approach or the other.

I am still not 100% on if we find WithinClauses defined why we might not have both types and therefore would need to account for that rather than one or the other only.

gfs:
If you have within clauses defined those will have narrowed down any matches to ones that match the conditions. 

If you check anything that's not a within clauses when you have any within clauses captures they will be A. Duplicates or B. ones that should have been filtered out by the within clause.

gfs:
I took a look at the code and I don't see anywhere that you define any rules with `RegexWithIndex` set as the custom operation.  That would be why it is not firing.  When did you expect it to fire?

ggjacosta:
I rechecked the docks and I only see the custom operation being defined along with a call to set the operation in the analyzer I don't see anything else which I did if there's something else that needs to be done while creating the rule or associating a rule to a custom operation I might need some example on how to do that.

gfs:
See this example: https://github.com/microsoft/OAT/wiki/Operations#delegate-example

Check the third code block where you set the custom operation field. And the fourth code block where you reference it in the created rule.

guyacosta:
Better -I am getting the custom op firing now and am getting the indexes after modifying the proposed solution to use the state1 as a text container similar to the scoped op previously and returning a tuple of int, boundary as I still want that not just a Match object so I think I'm on set for now and will complete the changes and a new commit.

guyacosta:
I think this should about do it.  It required adding the custom rule label to the rule conversation and some correction to the custom operation itself as mentioned but is working nicely and is now used instead of my less performant alternative version of pattern matching the clause to the rule patterns.  Additional changes include improvements to rule verification which now includes regex validation, language and other checks not present earlier causing in fact some rules to be fixed.  Manual and automated test are succeeding.  In a few cases, the counts diff from the previous version which I'm exploring further for a possible final commit to this branch but this is very close and provided a good opportunity to learn how to work with OAT better and I'm liking the delegate model.

guyacosta:
Small count of tag diffs mentioned earlier are addressed by last commit.

AppInspector/Commands/VerifyRulesCommand.cs
```diff
@@ -136,20 +136,9 @@ public VerifyRulesResult GetResult()
             try
             {
                 RulesVerifier verifier = new RulesVerifier(_rules_path, _options.Log);
-                verifier.Verify(_options.Failfast);
                 verifyRulesResult.ResultCode = VerifyRulesResult.ExitCode.Verified;
```
gfs, line 139: Do you want rules to be considered verified if the verification throws an exception?

RulesEngine/Ruleset.cs
```diff
@@ -81,217 +67,306 @@ public void AddDirectory(string path, string tag = null)
...
-            JsonSerializerSettings settings = new JsonSerializerSettings()
+            foreach (var rule in collection.Select(AppInspectorRuleToOatRule))
             {
-                Error = HandleDeserializationError
-            };
+                if (rule != null)
+                {
+                    if (_logger != null)
+                    {
+                        _logger.Debug("Attempting to add rule: " + rule.Name);
```
gfs, line 105: Replace this if statement and condition inside the statement with:

```suggestion
                        _logger?.Debug("Attempting to add rule: " + rule.Name);
```

RulesEngine/Ruleset.cs
```diff
@@ -81,217 +67,306 @@ public void AddDirectory(string path, string tag = null)
...
+        public ConvertedOatRule? AppInspectorRuleToOatRule(Rule rule)
+        {
+            if (rule == null)
+                return null;
 
-                        foreach (SearchCondition condition in r.Conditions)
-                        {
-                            SanitizePatternRegex(condition.Pattern);
-                        }
+            _rules.Add(rule); //add native Rule to avoid creating from enumerated _oatRules later
```
gfs, line 167: This is a pretty nasty side effect of calling this function.  I'd put this elsewhere if you want to keep a stash of the original rules.

gfs (commented):
Generally looks good!  From an OAT perspective this looks correct.

See comments.  Concerned about the function with side effects in RuleProcessor.

RulesEngine/Ruleset.cs
```diff
@@ -102,7 +102,7 @@ public void AddRange(IEnumerable<Rule> collection)
                 {
                     if (_logger != null)
```
gfs, line 103: You don't need this check with the ? on logger below.

gfs: The ? will, if the thing to the left is null, skip processing whats on the right and return null.  

For example when `_logger` is null.

With `_logger.Debug` you get a null reference error because you're trying to call `Debug` on `null`.

With `_logger?.Debug` instead of calling `Debug` the line just is a no-op.

gfs: If you have this `if (_logger is null)` pattern elsewhere you can replace it with the one line `?` version.

RulesEngine/Ruleset.cs
guyacosta: The counts always matched but maybe your thinking of the use of the yield in StringToRules so I moved it in a minimalist way to resolve.  Thanks.

gfs: The rules happened to match because you happened to only call this function once per rule. The function says "convertthis" it shouldn't have other side effects other than returning the converted object. Unless the function is "convert this and add it to a queue first" but then you need to change the name of the function. 

Imagine you want to test at some point and just want to call this function to convert. You never intended to change the state of the rules field you just wanted the converted object to test with.

gfs: Unless you access _rules extensively and absolutely need the performance (I don't think this is the case) my recommendation would be to ditch keeping track of them this way at all and use Linq to .Select the rules out when you want them. Just change _rules to be a property with 
```csharp
get{
    return _oatRules.Select(x => x.AppInspectorRule);
}
```

AppInspector/Commands/VerifyRulesCommand.cs
guyacosta: There's already an exception handler and the VerifyRulesResult is local so setting it's value for a critical event doesn't help.  Verified and NotVerified (non-critical i.e. file loads but Rule is not compliant) does return an error to either CLI or Nuget callers.  Both the CLI and Nuget callers are expected to try/catch critical errors.

RulesEngine/Ruleset.cs
```diff
@@ -277,14 +275,26 @@ public IEnumerable<ConvertedOatRule> ByLanguage(string language)
 
         public IEnumerable<ConvertedOatRule> GetOatRules() => _oatRules;
 
-        public IEnumerable<Rule> GetAppInspectorRules() => _rules;
+        public IEnumerable<Rule> GetAppInspectorRules()
```
gfs, line 278: This should work to replace this function and replace _rules so it works with the other functions.

```suggestion
        public IEnumerable<Rule> GetAppInspectorRules()
        {
            return _rules;
        }
        
        private IEnumerable<Rule> _rules { get =>_oatRules.Select(x => x.AppInspectorRule) }
```

RulesEngine/Ruleset.cs
guyacosta: very true...done elsewhere but will make consistent

gfs (changes requested):

AppInspector/Commands/VerifyRulesCommand.cs
gfs: So why set the result code at all?  You discard this value in all cases.  Just leave it the default enum value until you properly set it.

RulesEngine/Ruleset.cs
```diff
@@ -275,16 +258,10 @@ public IEnumerable<ConvertedOatRule> ByLanguage(string language)
 
         public IEnumerable<ConvertedOatRule> GetOatRules() => _oatRules;
 
+
```
gfs, line 261: Whitespace

gfs (approved):

Issue #278, opened 2020-09-02
Automatically highlight the line with finding, not the first line of the snippet


gfs:
For example, in the following the SerializeObject line is the found line, and should be highlighted.  Highlighting the first line is confusing as it draws attention to the context rather than the actual finding.

![image](https://user-images.githubusercontent.com/98900/91923853-352aed80-ec86-11ea-83ca-c91548fdd9c7.png)

guyacosta:
Fixed finally

Issue #259, opened 2020-08-06
Unique Tag Option Not Working Resulting in Duplicate Match Details


guyacosta:
When viewing a feature in the file dialog showing the code snippet and line for the sample value match the line number reflects only one of the files when more than one is listed.  In addition, the line numbers in this case start at 0.

Related -investigate whether there should be more than one file shown when the default "uniquetags" is used.

guyacosta:
The initial investigation was around why more than one file was appearing in the file dialog when the default unique tag options was true.  This lead to an addition to the RuleProcessor to capture results across files so that the BestMatch algorithm would work properly.  In addition, a simplification was made to remove a near redundant structure ScanResult in favor of the enhanced MatchRecord structure which is returned to the report anyway saving copy time.  Finally, the line count was corrected in case where it would appear as zero.

guyacosta:
fixed

PR #292, opened 2020-09-21
Fixes for #259, #278.  Includes correction for unique tags option tha…

guyacosta:
Fixes for #259, #278.  Includes correction for unique tags option that was not functioning right due to clause misidentification causing multiple files to be listed in the FileDialog.  Improves code by reducing near redundant ScanResult structure to use MatchResult directly and avoid duplication. Adds back 3 test cases post OAT integration.  Fixes line selection highlight.  Fixes AzureKeyVault pattern for regex-word.


guyacosta:
The test that is failing runs locally and succeeds.  The message may indicate a general failure to run the test starting with the first one.  I'll retry to see if it passes after an execution hickup.

Issue #298, opened 2020-09-23
MultiExtractor


yezz123:
MultiExtractor has been replaced using the now ported `Nuget` version RecursiveExtractor but several references still exist in the solution including the pipeline `YML` file

guyacosta:
Nice catch.  Will update to remove.

PR #301, opened 2020-09-27
Dependency package version updates

guyacosta:
#298 is also addressed to remove an outdated reference.


yezz123 (commented):
Effective changes!

PR #327, opened 2021-03-16
Add FileNamesRegex field to App inspector rules

gfs:


RulesEngine/Rule.cs
```diff
@@ -38,11 +41,40 @@ public class Rule
...
+            set 
+            { 
+                _updateCompiled = true; 
+                _appliesTo = value; 
+            } 
+        }
+
+        private string[]? _appliesTo;
+
+        [JsonProperty(PropertyName = "file_regexes")]
```
scovetta, line 60: Maybe `applies_to_file_regex`?

gfs: Sure. Changed.

RulesEngine/RuleProcessor.cs
```diff
@@ -216,28 +214,57 @@ void ProcessBoundary(ClauseCapture cap)
...
         /// <returns> List of rules </returns>
-        private IEnumerable<ConvertedOatRule> GetRulesForSingleLanguage(string language)
+        private IEnumerable<ConvertedOatRule> GetRulesByLanguage(string input)
+        {
+            string langid = string.Empty;
+
+            if (EnableCache)
+            {
+                // Make language id for cache purposes
+                langid = string.Join(":", input);
```
scovetta, line 232: I don't think this does anything.

RulesEngine/RuleProcessor.cs
```diff
@@ -216,28 +214,57 @@ void ProcessBoundary(ClauseCapture cap)
...
+        /// <returns> List of rules </returns>
+        private IEnumerable<ConvertedOatRule> GetRulesByFileName(string input)
         {
             string langid = string.Empty;
 
             if (EnableCache)
             {
                 // Make language id for cache purposes
-                langid = string.Join(":", language);
+                langid = string.Join(":", input);
```
scovetta, line 261: Same here, input is just a string.

scovetta (approved):

RulesEngine/RuleProcessor.cs
gfs:
```suggestion
                langid = string.Join(":", input);
```

This isn't actually new code. But yes, I don't know why that's being appended. I'll remove it.

RulesEngine/RuleProcessor.cs
gfs: Yup. I'll update it.

PR #335, opened 2021-03-30
Fix potential issue fetching keys

gfs:
user reports that toimmutablesorteddictionary can fail here.  We don't actually need the dictionary so changed to just grab the keys and sort them.

### DIFF
--- a/AppInspector/MetaDataHelper.cs
+++ b/AppInspector/MetaDataHelper.cs
@@ -139,17 +139,29 @@ namespace Microsoft.ApplicationInspector.Commands
         /// </summary>
         public void PrepareReport()
         {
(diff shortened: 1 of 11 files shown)


AppInspector/MetaDataHelper.cs
```diff
@@ -139,16 +139,28 @@ public void AddMatchRecord(MatchRecord matchRecord)
         /// </summary>
         public void PrepareReport()
         {
-            Metadata.CPUTargets = CPUTargets.ToImmutableSortedDictionary().Keys.ToList();
```
gfs, line 142: This is the only real change. The rest is just nullable safety.

jacobmsft (approved):

PR #340, opened 2021-04-05
Rewrite Analyze Command

### LINKED ISSUE #342: Per file timeout
Allow a per file processing timeout argument to prevent runaway analysis.

gfs:
* Refactor RuleProcessor.cs.
* Rewrites the Analyze Command for Performance and Reliability.  Removes some complex functionality and refactors into the Get Tags command.
* Adds the GetTags command to just get the unique tags that match.
* Adds a progress bar for analyze and get tags commands.
* Add GetTags tests and update test behavior to match new code.
* Fixes non-thread safe Match loss bug in Metadata.cs
* Adds Per File Timeout option (Fix #342) `--file-timeout N` to time out after N seconds.
* The metadata now reports on analysis status and duration for each file in Analyze and GetTags commands.
* The new paradigm is to do direct data collection in the threadsafe `MetaDataHelper` and then to compile it into a non-threadsafe but serializable `Metadata` report using the `CompileReport` command in `MetaDataHelper`.  Previously some data (like matches) was being directly collected in the metadata but some was being collected in the metadata helper.  If you see any cases that directly write data to the metadata that should no longer be the case.

Confirmed that the html report for single and multi-threaded shows the same number of results for a large project (.net codebase).

The number shown on the progress bar may differ but should not ultimately matter.

### DIFF
--- a/AppInspector.CLI/AppInspector.CLI.csproj
+++ b/AppInspector.CLI/AppInspector.CLI.csproj
@@ -65,8 +65,8 @@
   </ItemGroup>
   
   <ItemGroup>
-    <PackageReference Include="DotLiquid" Version="2.0.361" />
-    <PackageReference Include="NLog" Version="4.7.5" />
+    <PackageReference Include="DotLiquid" Version="2.1.418" />
+    <PackageReference Include="NLog" Version="4.7.9" />
   </ItemGroup>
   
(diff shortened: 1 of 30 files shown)


gfs:
See #344 instead to merge to dev branch.

PR #344, opened 2021-04-23
First Commit 1.3 Beta

### LINKED ISSUE #339: Rewrite Enumeration To be Individual Step
The first step should be enumerating all files into FileEntry objects.

This may be done in parallel with a reader that is processing those objects.

### LINKED ISSUE #341: Report Time taken for each file
In the metadata report how much time it took to process each file.

### LINKED ISSUE #342: Per file timeout

### LINKED ISSUE #343: Better handle recursive extractor overflow scenario.
Shouldn't result in a crash just a visible error to the user. Other files should be processed.

gfs:
* Refactor RuleProcessor.cs.
* Rewrites the Analyze Command for Performance and Reliability.  Removes some complex functionality and refactors into the Get Tags command.
* Adds the GetTags command to just get the unique tags that match.
* Adds a progress bar for analyze and get tags commands. (Fix #339)
* Add GetTags tests and update test behavior to match new code.
* Fixes non-thread safe Match loss bug in Metadata.cs
* Adds Per File Timeout option (Fix #342) `--file-timeout N` to time out after N milliseconds.
* The metadata now reports on analysis status and duration for each file in Analyze and GetTags commands. (Fix #341)
* The new paradigm is to do direct data collection in the threadsafe `MetaDataHelper` and then to compile it into a non-threadsafe but serializable `Metadata` report using the `CompileReport` command in `MetaDataHelper`.  Previously some data (like matches) was being directly collected in the metadata but some was being collected in the metadata helper.  If you see any cases that directly write data to the metadata that should no longer be the case.
* Skips files based on non printable character heuristic based on Perl's implementation.
* Update pipeline to build PRs to development.  We don't build releases from development.
* Fix #343 to better handle recursive extractor overflows.
* Provide cancellation functionality for Analyze and GetTags commands for use in time limited pipelines.

Confirmed that the html report for single and multi-threaded shows the same number of results for a large project (.net codebase).

The number shown on the progress bar may differ but should not ultimately matter.


AppInspector.CLI/CLICmdOptions.cs
```diff
@@ -19,6 +19,37 @@ public class CLICommandOptions : CommandOptions
...
+        [Option('c', "confidence-filters", Required = false, HelpText = "Output only matches with specified confidence <value>,<value> [high|medium|low]", Default = "high,medium")]
+        public string ConfidenceFilters { get; set; } = "high,medium";
+
+        [Option('k', "file-path-exclusions", Required = false, HelpText = "Exclude source files (none|default: sample,example,test,docs,lib,.vs,.git)", Default = "sample,example,test,docs,lib,.vs,.git")]
+        public string FilePathExclusions { get; set; } = "sample,example,test,docs,.vs,.git";
+
+        [Option("file-timeout", Required = false, HelpText = "Maximum amount of time in seconds to allow for processing each file.", Default = 0)]
+        public int FileTimeOut { get; set; } = 0;
+        
+        [Option("single-threaded", Required = false, HelpText = "Disables parallel processing.")]
```
scovetta, line 46: At some point, we should probably remove this option, I think?

AppInspector.CLI/CLICmdOptions.cs
```diff
@@ -19,6 +19,37 @@ public class CLICommandOptions : CommandOptions
...
+        [Option('r', "custom-rules-path", Required = false, HelpText = "Custom rules file or directory path")]
+        public string? CustomRulesPath { get; set; }
+
+        [Option('i', "ignore-default-rules", Required = false, HelpText = "Exclude default rules bundled with application", Default = false)]
+        public bool IgnoreDefaultRules { get; set; }
+
+        [Option('c', "confidence-filters", Required = false, HelpText = "Output only matches with specified confidence <value>,<value> [high|medium|low]", Default = "high,medium")]
+        public string ConfidenceFilters { get; set; } = "high,medium";
+
+        [Option('k', "file-path-exclusions", Required = false, HelpText = "Exclude source files (none|default: sample,example,test,docs,lib,.vs,.git)", Default = "sample,example,test,docs,lib,.vs,.git")]
```
scovetta, line 40: We should probably also exclude binary files -- e.g. .exe, .dll, .v2, .pdb, unless there's a nice fast way of maybe sampling the distribution of high ascii chars from the first 256 bytes of the file, or something like that?

Because I ran this against a basic VS source project and since the pdb, exe, etc. got pulled in, the resultant .html was huge.

AppInspector.CLI/CLICmdOptions.cs
```diff
@@ -31,15 +62,9 @@ public class CLIAnalyzeCmdOptions : CLICommandOptions
...
-        [Option('h', "match-depth", Required = false, HelpText = "First match or best match based on confidence level (first|best)", Default = "best")]
-        public string MatchDepth { get; set; } = "best";
-
         [Option('i', "ignore-default-rules", Required = false, HelpText = "Exclude default rules bundled with application", Default = false)]
         public bool IgnoreDefaultRules { get; set; }
 
-        [Option('d', "allow-dup-tags", Required = false, HelpText = "Output contains unique and non-unique tag matches", Default = false)]
-        public bool AllowDupTags { get; set; }
-
         [Option('b', "suppress-browser-open", Required = false, HelpText = "Suppress automatically opening HTML output using default browser", Default = false)]
```
scovetta, line 68: How do you feel about removing this entirely? It seems kinda spooky to have a CLI open a web browser by default. Users can just view the HTML file if they want.

AppInspector.CLI/CLICmdOptions.cs
```diff
@@ -58,8 +83,14 @@ public class CLIAnalyzeCmdOptions : CLICommandOptions
         [Option('t', "tag-output-only", Required = false, HelpText = "Output only identified tags", Default = false)]
         public bool SimpleTagsOnly { get; set; }
 
+        [Option("file-timeout", Required = false, HelpText = "Maximum amount of time in seconds to allow for processing each file.", Default = 0)]
+        public int FileTimeOut { get; set; } = 0;
```
scovetta, line 87: Better for this to be in milliseconds, maybe?

AppInspector.CLI/Program.cs
```diff
@@ -150,6 +152,16 @@ private static int VerifyOutputArgsRun(CLIPackRulesCmdOptions options)
             return RunPackRulesCommand(options);
         }
 
+        private static int VerifyOutputArgsRun(CLIGetTagsCommandOptions options)
+        {
+            Logger logger = Utils.SetupLogging(options, true);
+            WriteOnce.Log = logger;
```
scovetta, line 158: Separate bug, but any error logs are written to the Application Inspector install dir, not the current directory.

AppInspector/Commands/AnalyzeCommand.cs
```diff
@@ -275,467 +281,253 @@ private void ConfigRules()
...
-                                var streamByteArray = new byte[file.Content.Length];
-                                file.Content.Read(streamByteArray);
-                                ProcessInMemory(file.FullPath, Encoding.UTF8.GetString(streamByteArray), languageInfo);
-                            }
-                        }
-                        catch (Exception e)
-                        {
-                            WriteOnce.SafeLog($"Failed to parse {file.FullPath}. {e.GetType()}:{e.Message}", LogLevel.Info);
-                        }
+                        Thread.Sleep(10);
```
scovetta, line 462: Is there a better way of avoiding this?

scovetta (approved):
Overall, looks great. I **love** the progress bar too!

AppInspector.CLI/CLICmdOptions.cs
gfs: It's useful for low power systems and very useful for debugging.

AppInspector.CLI/CLICmdOptions.cs
gfs: The old behavior is to ignore everything that didn't have a known extension. Is that a better approach?

AppInspector.CLI/CLICmdOptions.cs
gfs: I'm fine with removing it.

AppInspector.CLI/CLICmdOptions.cs
gfs: Sure.

AppInspector/Commands/AnalyzeCommand.cs
gfs: What do you want to avoid? This isn't really stalling it's just looping to write the number of results to the screen.

AppInspector.CLI/CLICmdOptions.cs
scovetta: Hmm, we had to do TreatEverythingAsCode in order to get it to scan everything. Not sure what the best approach is, but with this PR, if you scan a directory with a .pdb in it, you'll get gobbly-gook in the html assuming some text matches a rule.

AppInspector/Commands/AnalyzeCommand.cs
scovetta: Ahh, my mistake then! :-)

AppInspector.CLI/CLICmdOptions.cs
gfs: I could only scan files that are valid UTF8?

AppInspector.CLI/Program.cs
gfs: Is this just for GetTags or in the analyze command too?  Install dir or run dir?

AppInspector.CLI/CLICmdOptions.cs
gfs: I added a check to look at the first 100 characters and if more than 50% of them are control characters it will skip the file.

AppInspector.CLI/Program.cs
gfs: I see now. Fixed.

AppInspector.CLI/CLICmdOptions.cs
gfs: Checking a handful of binaries shows they're over 50% of the first 1024 characters, in particular a lot of \0 and this check works.  Open to any thoughts on tweaking how many characters to check and how many should be control as well.  50% may actually be a very high proportion. There shouldn't be many if any non-printable characters in source code.  Haven't checked the performance implications but could also check every character in the file (the streams should be in memory most of the time), this likely isn't that much extra processing to scan the file one more time if you consider how many regexes we run on the entire file.

gfs: Rewrote this to match Perl's behavior which is "Is binary if it has any NULs or if its more than 30% non -printable characters"

AppInspector.CLI/CLICmdOptions.cs
```diff
@@ -22,26 +22,48 @@ public class CLICommandOptions : CommandOptions
     /// <summary>
     /// CLI command distinct arguments
     /// </summary>
-    [Verb("analyze", HelpText = "Inspect source directory/file/compressed file (.tgz|zip) against defined characteristics")]
-    public class CLIAnalyzeCmdOptions : CLICommandOptions
+    [Verb("get-tags", HelpText = "Inspect source directory/file/compressed file (.tgz|zip) against defined characteristics for unique tags")]
```
jacobmsft, line 25: are there any other tools other than codeinspector using application inspector for automation? If yes, this change will break it when they upgrade, if not nevermind.

AppInspector.CLI/Program.cs
```diff
@@ -263,6 +263,31 @@ private static void ValidFileWritePath(string filePath)
...
+        {
+            GetTagsResult.ExitCode exitCode = GetTagsResult.ExitCode.CriticalError;
+
+            GetTagsCommand command = new GetTagsCommand(new GetTagsCommandOptions()
+            {
+                SourcePath = cliOptions.SourcePath ?? "",
+                CustomRulesPath = cliOptions.CustomRulesPath ?? "",
+                IgnoreDefaultRules = cliOptions.IgnoreDefaultRules,
+                ConfidenceFilters = cliOptions.ConfidenceFilters,
+                FilePathExclusions = cliOptions.FilePathExclusions,
```
jacobmsft, line 276: Does FilePathExclusions include extension filters? Current tool has an option to filter out specific extensions

AppInspector.CLI/ResultsWriter.cs
```diff
@@ -58,20 +58,19 @@ public static void Write(Result result, CLICommandOptions options)
                     writer?.WriteResults(analyzeResult, cLIAnalyzeCmdOptions);
 
                     //post checks
-                    if (File.Exists(options.OutputFilePath) && new FileInfo(options.OutputFilePath).Length > MAX_HTML_REPORT_FILE_SIZE)
+                    if (options.OutputFilePath is not null && File.Exists(options.OutputFilePath) && new FileInfo(options.OutputFilePath).Length > MAX_HTML_REPORT_FILE_SIZE)
```
jacobmsft, line 61: This is the output file path, right? What if the file doesn't exist?  Is it created before this line?
File.Exists(options.OutputFilePath)

AppInspector.CLI/Writers/AnalyzeHtmlWriter.cs
```diff
@@ -56,23 +55,23 @@ private void WriteHtmlResult()
             RenderResultsSafeforHTML();
 
             //Grab any local css and js files that are needed i.e. don't have hosted URL's or are proprietary
-            string allCSS = "<style>\n" + MergeResourceFiles(Path.Combine(Utils.GetPath(Utils.AppPath.basePath), "html","resources","css"));
-            string allJS = "<script type=\"text/javascript\">\n" + MergeResourceFiles(Path.Combine(Utils.GetPath(Utils.AppPath.basePath),"html","resources","js"));
+            string allCSS = "<style>\n" + MergeResourceFiles(Path.Combine(Utils.GetPath(Utils.AppPath.basePath), "html", "resources", "css"));
+            string allJS = "<script type=\"text/javascript\">\n" + MergeResourceFiles(Path.Combine(Utils.GetPath(Utils.AppPath.basePath), "html", "resources", "js"));
```
jacobmsft, line 59: nit; but "include everything" is a good target for supply chain attacks.

AppInspector.CLI/Writers/WriterFactory.cs
```diff
@@ -111,7 +115,37 @@ public static CommandResultsWriter GetExportWriter(CLIExportTagsCmdOptions optio
...
+            return writer;
+        }
+
+        private static CommandResultsWriter GetGetTagsWriter(CLIGetTagsCommandOptions options)
+        {
+            CommandResultsWriter? writer;
+
+            switch (options.OutputFileFormat.ToLower())
+            {
+                case "_dummy":
```
jacobmsft, line 139: is this for testing?

AppInspector.CLI/CLICmdOptions.cs
gfs: Why would this break them? The analyze command still exists.

AppInspector.CLI/Program.cs
gfs: File path exclusions does a .contains in the things listed as arguments. I didn't change this behavior.

AppInspector.CLI/Writers/WriterFactory.cs
gfs: I don't know what it's used for. I am just following the existing patterns.

AppInspector.CLI/Writers/AnalyzeHtmlWriter.cs
gfs: Agreed but I think I'll put off any modifications to the html output til later.

AppInspector/Commands/AnalyzeCommand.cs
```diff
@@ -275,467 +255,291 @@ private void ConfigRules()
...
-                        _metaDataHelper?.Metadata.IncrementFilesSkipped();
-                        return;
-                    }
+                    if (cancellationToken.IsCancellationRequested) { break; }
+                    ProcessAndAddToMetadata(entry);
+                }
+            }
+            else
+            {
+                Parallel.ForEach(populatedEntries, new ParallelOptions() { CancellationToken = cancellationToken }, entry => ProcessAndAddToMetadata(entry));
```
jacobmsft, line 299: This is voluntary task completing with a cancellation token. I dont know if it makes sense to do a preemptive one when the max allowed run time (an universal timer when it expires, exit process). What do you think?

AppInspector.CLI/ResultsWriter.cs
gfs: Line 58 creates the file.

AppInspector.CLI/Program.cs
gfs: I don't believe there was ever an explicit extension filters. I didn't remove it. 

If you'd like one instead of or in addition to this I'm happy to change how this behaves. It could be regexes instead for example.

AppInspector/Commands/AnalyzeCommand.cs
gfs: Fixed not catching the operation cancelled exception here. As discussed should now behave as expected.

AppInspector.CLI/Writers/WriterFactory.cs
gfs: I looked more into it.  These just throw exceptions. I'm going to get rid of the dummy writers.

AppInspector/Commands/AnalyzeCommand.cs
gfs: This will throw here which will be caught and get result will return Canceled as the status.

AppInspector.CLI/CLICmdOptions.cs
gfs: To clarify there are some changes to the flags that might break them (only removal no flags have changed meaning).  I don't think there is any actual missing functionality net for net though.  You can use the GetTags command instead of the old AnalyzeCommand with the arguments to limit down to single matches.

PR #365, opened 2021-05-12
Performance and misc improvements

### LINKED ISSUE #362: Add OAT Contains with index operator for performance
Improve performance by using string matches as .indexof operations returning boundaries instead of converting to regex which should improve performance somewhat.

gfs:
Switch to using indexof instead of regex when possible.  Showing significant performance uplift.

Fix #362.

### DIFF
--- a/RulesEngine/OatSubstringIndexClause.cs
+++ b/RulesEngine/OatSubstringIndexClause.cs
@@ -1,0 +1,20 @@
+﻿// Copyright (C) Microsoft. All rights reserved. Licensed under the MIT License.
+
+using Microsoft.CST.OAT;
+
+namespace Microsoft.ApplicationInspector.RulesEngine
+{
+    public class OATSubstringIndexClause : Clause
+    {
+        public OATSubstringIndexClause(PatternScope[] scopes, string? field = null, bool useWordBoundaries = false) : base(Operation.Custom, field)
+        {
+            Scopes = scopes;
+            CustomOperation = "SubstringIndex";
+            UseWordBoundaries = useWordBoundaries;
+        }
+
+        public PatternScope[] Scopes { get; }
+
+        public bool UseWordBoundaries {get;}
+    }
+}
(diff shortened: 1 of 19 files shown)


RulesEngine/OatSubstringIndexClause.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (C) Microsoft. All rights reserved. Licensed under the MIT License.
+
+using Microsoft.CST.OAT;
```
jacobmsft, line 3: Is OatSubstringIIndex generic enough to move into OAT: https://github.com/microsoft/OAT/tree/main/OAT/Operations?, or did I get that wrong?

RulesEngine/OatSubstringIndexOperation.cs
```diff
@@ -0,0 +1,101 @@
...
+            CustomOperation = "SubstringIndex";
+            OperationDelegate = SubstringIndexOperationDelegate;
+            ValidationDelegate = SubstringIndexValidationDelegate;
+        }
+
+        public IEnumerable<Violation> SubstringIndexValidationDelegate(CST.OAT.Rule rule, Clause clause)
+        {
+            if (clause.Data?.Count == null || clause.Data?.Count == 0)
+            {
+                yield return new Violation(string.Format(Strings.Get("Err_ClauseNoData"), rule.Name, clause.Label ?? rule.Clauses.IndexOf(clause).ToString(CultureInfo.InvariantCulture)), rule, clause);
```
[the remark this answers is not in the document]
gfs: The api you implement is an enumerable so
1. You can Fail fast when validating on the first error if you don't care what they all are 
2. The function doesn't have to keep a list of the found errors.

RulesEngine/OatSubstringIndexClause.cs
gfs: A few issues. 

It's not much different than the contains operator that already exists and is much more broad in terms of the data types it can work on. 

Secondly, built-in operations in OAT should not be for one specific data type (for example this just works for strings).

The main distinction is that we are capturing and returning the index of the match, a behavior that is not mirrored in any other OAT operation, and doesn't make contextual sense outside of simple string matches.

See the existing AI code and AI uses it's own regex instead of the OAT regex because OAT doesn't deal with Boundary objects, that's a AI concept.

PR #369, opened 2021-05-13
Cache Results of IsCommented

### LINKED ISSUE #370: "regex-word" rule patterns have the wrong precedence
**Describe the bug**
Example:
Take a rule like AI015600 (Data: Audio File). The raw pattern is `mp2|mp3|mp4...` with a type of `regex-word`. This translates to the regex `\bmp2|mp3|mp4...\b`, which isn't right -- e.g. `foomp3bar` matches this, since the `\b`'s are attached to the first and last strings.

It looks like we have a bunch of this -- it might be a regression in the dev branch, but we might want to update:

https://github.com/microsoft/ApplicationInspector/blob/development/RulesEngine/Ruleset.cs#L401

from:
`pattern.Pattern = string.Format(CultureInfo.InvariantCulture, @"\b{0}\b", pattern.Pattern);`
to
`pattern.Pattern = string.Format(CultureInfo.InvariantCulture, @"\b({0})\b", pattern.Pattern);`

**To Reproduce**
1. Download NPM/left-pad@1.3.0.
2. Run Application Inspector (latest beta) on it.
3. Notice the false positives, like AI015600 found in package/perf/perf.js (matching "hma" in "Benchmark").

**Expected behavior**
Regex-word should only match complete words (e.g. `\b(THING)\b`).

**Screenshots**
n/a

**Additional context**
ApplicationInspector.CLI 1.3.11-beta+69410edbd6

gfs:
Cache the results of IsCommented and simplify algorithm to improve performance.
Fix GetLineBoundary.
Update tests.
Update rules.
Remove TagTest command.
Rewrite TagDiff Command.
Use Globs for file path exclusions.
Change default file path exclusions.
Set a default file timeout of 60 seconds.
Can now provide multiple paths to scan at a time comma separated.
Limit number of threads when running parallel to 1.5 * processor count.
"regex-word" is now "regexword". 

Fix #370


daalcant (approved):

AppInspector/Commands/AnalyzeCommand.cs
```diff
@@ -333,17 +344,17 @@ void ProcessAndAddToMetadata(FileEntry file)
                     }
                     else
                     {
-                        _ = _metaDataHelper?.FileExtensions.TryAdd(Path.GetExtension(file.FullPath).Replace('.', ' ').TrimStart(), 0);
```
jacobmsft, line 336: .ToLower().Replace('.', ' ').TrimStart()?

AppInspector/Commands/AnalyzeCommand.cs
```diff
@@ -319,7 +330,7 @@ void ProcessAndAddToMetadata(FileEntry file)
                 var sw = new Stopwatch();
                 sw.Start();
 
-                if (_fileExclusionList.Any(x => file.FullPath.ToLower().Contains(x)))
+                if (_fileExclusionList.Any(x => x.IsMatch(file.FullPath)))
```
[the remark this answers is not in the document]
jacobmsft: You might have to pass the matcher around to filter zip files though.

AppInspector/Commands/AnalyzeCommand.cs
```diff
@@ -662,30 +737,31 @@ public AnalyzeResult GetResult()
 
                     while (!done)
                     {
-                        Thread.Sleep(10);
-                        var current = _metaDataHelper?.Files.Count ?? 0;
+                        Thread.Sleep(100);
```
jacobmsft, line 740: 100 seems to be repeated in a couple places, so maybe a constant?

AppInspector/Commands/AnalyzeCommand.cs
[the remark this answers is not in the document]
jacobmsft: but this would not give files with ScanState.Skipped status

AppInspector/Commands/TagDiffCommand.cs
```diff
@@ -11,12 +11,14 @@ namespace Microsoft.ApplicationInspector.Commands
 {
     public class TagDiffOptions : CommandOptions
```
jacobmsft, line 12: This doesn't seem to be exactly TagDiffs, since TagDiffs would contain added and removed tags. More like newTags/TagsAdded?

jacobmsft (approved):
There seems to be dlls checked in...should these be removed?

AppInspector/Commands/AnalyzeCommand.cs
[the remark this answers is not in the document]
gfs: I like keeping it here so we can have the scan state skipped.  I think I'll stick with how it is now.

AppInspector/Commands/AnalyzeCommand.cs
gfs: Good idea.

AppInspector/Commands/TagDiffCommand.cs
gfs: Thanks for pointing this out.  THis is supposed to create the diffs.  I've fixed it.

AppInspector/Commands/AnalyzeCommand.cs
gfs: This is a removed line.

PR #381, opened 2021-06-03
Clean up existing rules

gfs:

PR #382, opened 2021-06-08
Remove extra ',' in rule file causing json parsing issues

jacobmsft:


gfs:
good catch, I incorporated this into #381

PR #394, opened 2021-09-06
Comments : Ruby : Skip code under multi-line commented portion

hiteshbedre:
To be able to skip the scanning between the commented code written in Ruby I mentioned prefix and suffix commenting style for the same. This change will just skip the matches/samples found under ```"=begin"``` and ```"=end"```.

### DIFF
--- a/RulesEngine/Resources/comments.json
+++ b/RulesEngine/Resources/comments.json
@@ -27,7 +27,6 @@
       "perl6",
       "r",
       "shellscript",
-      "ruby",
       "yaml",
       "powershell",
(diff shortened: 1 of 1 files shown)


gfs:
Ruby multiline comments appear to be unique in that they must begin at the start of a line. If a multiline comment is started after some code on a line it will actually be a syntax error. I don't think it is worth at this time implementing that exact behavior to only match multi line comments on line start as 1. as it would require a number of changes across the code base so this seems like a good enough solution 2. its not valid ruby anyway.

Issue #396, opened 2021-11-02
TextContainer GetLocation Throws Exception Occasionally 


guyacosta:
**Describe the bug**
On occasion there the RulesEngine throws an exception for certain files when trying to determine match locations from the TextContainer.cs file indicated.  The issue appears to be a conditional that allows there to be an uneven number of items between the LineEnds and LineStarts lists culminating in a failure in the GetLocation method on line 218.  The solution appears to be correcting the boundary limit in the constructor to include '<=' on line 35.

**To Reproduce**
Source code to reproduce will vary...was primarily python if that helps.
The behavior interestingly was seen while using the NuGet model but not the commandLine one -not sure why but time is limited to investigate but it is also not consistent for reasons unknown i.e. sometimes it doesn't throw an exception.

**Expected behavior**
Complete analysis for source without throwing an exception

**Code Section**
..\ApplicationInspector-main\RulesEngine\TextContainer.cs
Class & Method: TextContainer : TextContainer
Line: 35
```
while (pos > -1)
{
           **if (pos > 0 && pos + 1 < FullContent.Length)**
           {
                    LineStarts.Add(pos + 1);
           }
           pos = FullContent.IndexOf('\n', pos + 1);
     
           if (LineEnds.Count < LineStarts.Count)
           {
                LineEnds.Add(FullContent.Length - 1);
           }
}
```
and in method GetLocation line 218 where the indexing exception gets thrown
```
result.Column = index - LineStarts[i];
```     

**Screenshots**
![appinsbug-txtcontainer](https://user-images.githubusercontent.com/47648296/139959534-439a5e3e-d91f-46bc-91cf-c1af7bb412ef.png)

**Additional context**

gfs:
> The solution appears to be correcting the boundary limit in the constructor to include '<=' on line 35.

https://github.com/microsoft/ApplicationInspector/blob/0e2ea0c2ebba23dd153130191527cf66acad009e/RulesEngine/TextContainer.cs#L26-L45

Have you tested that change to confirm it fixes this issue? The 'pos' variable is 0 indexed but it is being compared to a `Length` which is essentially 1 indexed.  The check for `pos + 1 < Length` ensures that we don't add a Line Start after the file has ended, a position equal to length is beyond the length of the 0 indexed string.

I do think the issue is on this line though. I have reproduced that if the first character of your file is `\n` then it will skip that LineStart. I haven't gotten it to crash from this, but it does desync the line starts and ends.

[...]

PR #398, opened 2021-11-03
Fix addition for #396

guyacosta:
Initializes Location values and adds safe return in the event of an index failure which is less likely now.


gfs:
This causes a new test failure resulting in a different number of results. Will have to evaluate.

I'm not sure that returning an incorrect position is better than an exception, I think null might be more appropriate.

gfs:
Will not merge this change.

gfs:
I don't think silently throwing this exception away is correct.

The underlying issue should have been resolved by the previous fix, doing an empty catch here will just prevent us from detecting any similar problems in the future. Returning 1,1 is incorrect if the position cannot be found.

PR #434, opened 2022-02-22
Allow users to provide their own languages and comments styles

### LINKED ISSUE #418: Add custom extensions in analysis execution.
**Is your feature request related to a problem? Please describe.**
We want to add some languages to https://github.com/microsoft/ApplicationInspector/blob/main/RulesEngine/Resources/languages.json list, like css or .md files, in order to reduce the number of unknown file types in our projects.
We can make a PR if the languages are worldwide known (like css or markdown), but we think it might be interesting if the CLI allows to add a custom json file with the same format with more file extensions in executions with a new cli parameter.

gfs:
Fix #418


gfs:
Won't merge this PR. See #441

Issue #443, opened 2022-04-27
HTML Report: File listing line number does not update


scovetta:
**Describe the bug**
When you click a new file to view the contents of (via Summary / click icon), the "Line Number: 42" on the bottom of the modal doesn't update.

**To Reproduce**
1. Default HTML report, view summary, click icon, click finding tag, change to a different file.

**Expected behavior**
Line number should update when a new file is clicked.

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Edge
 - Version: [687cf00](https://github.com/microsoft/ApplicationInspector/commit/687cf00e8496a08e4ea045c0e73294531428ed9e)

gfs:
Fixed in 1.5-beta.

Issue #442, opened 2022-04-27
Application Inspector HTML tags aren't showing properly


scovetta:
**Describe the bug**
All of the tag icons are highlighted, but none are clickable. There is a JS error saying that `match.tags` is not defined.

**To Reproduce**
1. Run the default CLI / analyze against a directory.
2. Open output.html
3. Click Summary.

**Expected behavior**
Icons should be clickable.

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Edge
 - Version: [687cf00](https://github.com/microsoft/ApplicationInspector/commit/687cf00e8496a08e4ea045c0e73294531428ed9e)

**Additional context**

gfs:
Fixed in 1.5-beta.

PR #444, opened 2022-04-27
Fixes for #442 and #443

scovetta:
This PR contains cosmetic fixes to the HTML report, the most important of which is because the match object's keys are now Uppercase, and the appinspector.js file had references to them all in lowercase, so it was causing the UI to break.


gfs:
Nice catch Mike.

gfs (approved):

PR #462, opened 2022-06-22
Update bug report template.

scovetta:
Minor fixes to the bug report template.


gfs (commented):
The examples use "appinspector" as that's the command to access the nuget tool when installed. Maybe I can add example for each.

Agree with the first line change. 

Will address tomorrow.

gfs:
I just removed the specific mention of AppInspector.CLI/AppInspector. We only need the arguments in the report in any case.

PR #466, opened 2022-06-22
Use Newtonsoft for all json operations

### LINKED ISSUE #465: Too much data is serialized to the json report
**Describe the bug**
Some fields appear to be being serialized which are not needed. I'm saw some line ends fields in a report presumably from a text container. Text container should not be serialized as part of the report. 

Full scope not yet known.

gfs:
There was inadvertently a mix of System.Text.Json and newtonsoft being used which resulted in undesired behavior like serializing the TextContainer object referenced in the MatchRecord object due to a mismatch between System.Text.Json's JsonIgnore and Newtonsoft's JsonIgnore.

Fix #465.

### DIFF
--- a/AppInspector.CLI/html/resources/js/appinspector.js
+++ b/AppInspector.CLI/html/resources/js/appinspector.js
@@ -134,12 +134,12 @@ class TemplateInsertion {
             }
         });
 
-        $('#s_pi_application_name').html(this.mt.applicationName);
-        $('#s_pi_version').html(this.mt.sourceVersion);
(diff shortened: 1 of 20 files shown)


gfs:
Huh. This is pretty strange to me because the camel casing is actually done by JsonPropertyName attributes on the classes that existed in 1.4 as well. It looks like this was a change made in deef493 to make them pascalcase. I think we likely detected that the report broke because of the json serializer mixing but we didn't understand the implications/root cause at the time. 

They "should" be camel case to be consistent with past behavior.

gfs:
Merged Mike's branch instead.

PR #467, opened 2022-06-22
Switch match references to camelCase.

scovetta:
It looks like recent changes for 1.5.x have the match objects as camelCase, where the references in appinspector.js are for PascalCase. This PR changes the JS references to camelCase -- feel free to change on the other end and ignore this PR.


gfs:
LGTM.

Issue #420, opened 2022-01-18
Pattern type for structured data files (yaml, json...)


valinha:
**Is your feature request related to a problem? Please describe.**
In order to be able to detect feature usage in structured configuration files (yaml, json, xml...) it would be very useful to be able to search for occurrences of certain entries regardless of the order of declaration in the file. 

An example would be:

If we have a config.yaml file and we have the following configuration:
``` yaml 
app:
  myfeature:
     prop1: x
     prop2: x
     propN: x
     enabled: true
```

If I want to detect with a rule if the 'myfeature' functionality is enabled I would not have a reliable solution at the moment. Detecting if app.my-feature.enabled = true with patterns, modiffiers or conditions does not cover all cases, e.g. if the declaration order is changed or the configuration is put in one line. I.e. a pattern such as:

"pattern": "app:\s+myfeature:\s+enabled: *true".

It would not cover the above example since the order of the enabled property can be changed as it is a structured data type.

**Describe the solution you'd like**
Ability to be able to perform patterns on structured data types.

**Describe alternatives you've considered**

- Implement new pattern types to support operations on structured data types (yaml, json...).
- Provide an extension mechanism to be able to do file processing with parsing tools.

gfs:
Thanks for the suggestion.  This is non trivial but I think its doable.

Here's a first thought about how we might be able to specify that from a user perspective in the rule object without changing the schema too much.

If path is specified, then if the file can be parsed as structured data (XML, JSON, YML), find all the paths in the document that match the path, and run the pattern against their value.

1. This doesn't change the operations you can perform on the end value, so string, regex etc would remain available.
2. It should be possible to do paths with wildcard parents, but I'm not sure about using wildcards in the path itself.

Here's the rule you want above implemented in this schema:

```json
{
    "name": "My Feature",
    "id": "MYID00001",
    "description": "Detects if My Feature is enabled",
    "applies_to": [
      "json"
    ],
    "tags": [
      "Features.MyFeature"
    ],
    "severity": "Moderate",
    "patterns": [
      {
        "confidence": "High",
        "pattern": "true",
        "type": "string"
      },
    "path": "app.myfeature.enabled"
    ]
  }
```

What do you think of this proposal?

valinha:
Thanks for considering the functionality. I'm fine with it, if it's true that I would like it not to be dependent on the applies_to field, something like that would work as well?

...
"applies_to_file_regex": [
  "application-?.*.yml".
],
...

[...]

PR #499, opened 2022-08-11
Workaround for namespace based parsing.

### LINKED ISSUE #497: Support XPath in XML Documents with Namespaces Specified
**Is your feature request related to a problem? Please describe.**
The current implementation of the XPath queries from #420 doesn't support XML documents which specify a namespace. 

**Describe the solution you'd like**
Options:
1. Somehow automatically parse the namespaces from the XML
2. Add a field for users to specify namespaces to use for the query
3. Strip the namespaces and treat elements as local
 
**Additional context**
See #420 for samples which work without namespace and fail to work with namespace.  I've done some quick experiments but have not cracked how to properly associate the namespace.  Once namespaces are supported they will also have to be specified when relevant in the xpath query itself. Might also require some better error handling around loading namespaces and queries.

gfs:
Fix #497

This makes some changes to the XML parsing so you can use the xpath `local-name` method.

Also adds a test using the below rule and sample xml to ensure matching works the same both with and without namespace when using local-name. 

This rule will match the following sample file now.
```json
[{
    "name": "Source code: Java 17",
    "id": "CODEJAVA000000",
    "description": "Java 17 maven configuration",
    "applies_to_file_regex": [
      "pom.xml"
    ],
    "tags": [
      "Code.Java.17"
    ],
    "severity": "critical",
    "patterns": [
      {
        "pattern": "17",
        "xpaths" : ["/*[local-name(.)='project']/*[local-name(.)='properties']/*[local-name(.)='java.version']"],
        "type": "regex",
        "scopes": [
          "code"
        ],
        "modifiers": [
          "i"
        ],
        "confidence": "high"
      }
    ]
  }]
```
Content reported as not working from #420 which now works:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>

  <groupId>xxx</groupId>
  <artifactId>xxx</artifactId>
  <version>0.1.0-SNAPSHOT</version>
  <packaging>pom</packaging>

  <name>${project.groupId}:${project.artifactId}</name>
  <description />

  <properties>
    <java.version>17</java.version>
  </properties>

</project>
```

### DIFF
--- a/AppInspector.RulesEngine/TextContainer.cs
+++ b/AppInspector.RulesEngine/TextContainer.cs
@@ -148,8 +148,13 @@ namespace Microsoft.ApplicationInspector.RulesEngine
                 {
                     if (nodeIter.Current is not null)
                     {
-                        var outerLoc = FullContent[minIndex..].IndexOf(nodeIter.Current.OuterXml);
-                        var offset = FullContent[outerLoc..].IndexOf(nodeIter.Current.InnerXml) + outerLoc + minIndex;
+                        // First we find the name
+                        var nameIndex = FullContent[minIndex..].IndexOf(nodeIter.Current.Name);
+                        // Then we grab the index of the end of this tag.
+                        // We can't use OuterXML because the parser will inject the namespace if present into the OuterXML so it doesn't match the original text.
+                        var endTagIndex = FullContent[nameIndex..].IndexOf('>');
+                        var totalOffset = nameIndex + endTagIndex + minIndex;
+                        var offset = FullContent[totalOffset..].IndexOf(nodeIter.Current.InnerXml) + totalOffset;
                         // Move the minimum index up in case there are multiple instances of identical OuterXML
                         // This ensures we won't re-find the same one
                         minIndex = offset;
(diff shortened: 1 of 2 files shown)


valinha:
I have tried the same but in the html report I don't see any match.

image

gfs:
Hmm. That's odd. I'll take a look.

gfs:
The HTML report seems to work fine for me.  Did you include add `-A` to allow all tags in 'build' files?

I installed the latest beta and ran `appinspector analyze -s pom.xml -A -i -r pomrule.json` using the xml data and json from above and it is finding the tag:

image

gfs:
Tried the same with pom.xml as applies to and it works as well - so updated the example rule above.

gfs:
@valinha are you still missing detections if you include -A?

Issue #510, opened 2022-09-30
Support .NET 7 Non-Backtracking Regex as option


gfs:
Provide an option flag on the RuleProcessor to use non backtracking regex.

https://devblogs.microsoft.com/dotnet/regular-expression-improvements-in-dotnet-7/#backtracking-and-regexoptions-nonbacktracking

dbalikhin:
I'm trying to understand how analyzer actually works and it simply goes to OAT project. 
OAT does actual regex search. It has a couple of parameters: `i` and `m`.
https://github.com/microsoft/OAT/blob/main/OAT/Operations/RegexOperation.cs#L51

It looks like we can set these params through pattern's modifiers (at least I can see `i` here).
    "patterns": [
      {
        "pattern": "authorize|authorized|authorization|isauthorized",
        "type": "regexword",
        "scopes": [
          "code"
        ],
        "modifiers": [
          "i"
      ...
      }

Are you talking about supporting a new modifier to let OAT know to use nonbacktracking?
Or it is a parameter to CLI?
Or both?

gfs:
So application inspector uses OAT to execute rules - but the operation delegates are defined in the app inspector code base, I don't believe this change requires changing OAT. 

See this for the regex operation:
https://github.com/microsoft/ApplicationInspector/blob/505942298a290aad129c6e26c6b748c80b94ecf5/AppInspector.RulesEngine/OatExtensions/OatRegexWithIndexOperation.cs#L80

This would require changing the method that converts the app inspector deserialized rule to oat rule (below) and then modifying the operation delegate from above to respect the new parameter. 

https://github.com/microsoft/ApplicationInspector/blob/505942298a290aad129c6e26c6b748c80b94ecf5/AppInspector.RulesEngine/AbstractRuleSet.cs#L75

gfs:
That said. I thought of this option making more sense as a parameter to the cli/lib to apply to all patterns used in a run - rather than something specified in a per pattern basis. So how to do that may require a bit more thought as to how to get that parameter to the right place for when the regexes are instantiated.

gfs:
One way to do this could be to add an option to the analyze options and cli analyze options to enable nonbacktracking and pass that to the rule set or have a property to set in the abstract rule set to cause the oat rule generation to pass the non backtracking parameter. 

Another way would be to, after the rules are deserialized to app inspector rules but before they are converted to oat rules inject the option there and then have the conversion method check for that option as well as `i` and `m`.

[...]

PR #528, opened 2023-02-14
Support non-backtracking regex

dbalikhin:
Fixes: #510 

Opening a draft request.
Testing + tests + sanity checks remaining.

Summary:
New EnableNonBacktrackingRegex property that comes from CLI input. False by default.
During rule conversion to OAT, it will check for this flag and add `nb` modifier unless `b` is already specified in the rule.
Thus, a local `b` rule modifier will overwrite global non-backtracking setting (if we want to use NonBacktracking but some rules require backtracking).
RuleVerifier will try with NB and if it fails, try with Backtracking.


AppInspector.RulesEngine/OatExtensions/OatRegexWithIndexOperation.cs
```diff
@@ -86,6 +86,14 @@ private static IEnumerable<Violation> RegexWithIndexValidationDelegate(CST.OAT.R
             RegexOptions regexOpts = Common.Utils.RegexModifierToRegexOptions(src.Arguments);
             var subBoundary = state2 is Boundary s2 ? s2 : null;
 
+#if NET7_0_OR_GREATER
```
gfs, line 89: Im not sure this is the right place to add this. The validation delegate isn't guaranteed to run. Should likely be in the RegexModiferToRegexOptipns method instead.

AppInspector.RulesEngine/RulesVerifier.cs
```diff
@@ -148,12 +148,27 @@ public RuleStatus CheckIntegrity(ConvertedOatRule convertedOatRule)
                     {
                         throw new ArgumentException();
                     }
-
+                    
                     RegexOptions regexOpts =
-                        Utils.RegexModifierToRegexOptions(searchPattern.Modifiers ?? Array.Empty<string>());
+                        Utils.RegexModifierToRegexOptions(searchPattern.Modifiers);
+                    
+#if NET7_0_OR_GREATER
```
gfs, line 155: Same as above. This should be part of the conversion method.

AppInspector.RulesEngine/RulesVerifier.cs
```diff
@@ -148,12 +148,27 @@ public RuleStatus CheckIntegrity(ConvertedOatRule convertedOatRule)
...
 
                     _ = new Regex(searchPattern.Pattern, regexOpts);
                 }
+                catch (NotSupportedException e)
+                {
+                    // Regex requires backtracking, try without non-backtracking
+                    _logger?.LogWarning(MsgHelp.GetString(MsgHelp.ID.VERIFY_RULES_REGEX_FAIL), rule.Id ?? "", searchPattern.Pattern ?? "", e.Message);
+                    errors.Add(MsgHelp.FormatString(MsgHelp.ID.VERIFY_RULES_REGEX_FAIL, rule.Id ?? "", searchPattern.Pattern ?? "", e.Message));
+
+                    _ = new Regex(searchPattern.Pattern, Utils.RegexModifierToRegexOptions(searchPattern.Modifiers));
```
gfs, line 170: Should use regexOpts from before with

`regexOpts &= ~RegexOptions.NonBackTracking` right?

gfs:
Will review more thoroughly this afternoon.

gfs (changes requested):
Will do a more thorough review this afternoon. But a few things stood out to me here.

AppInspector.RulesEngine/RulesVerifier.cs
dbalikhin: regexOpts won't be available here due to the declaration scope. We will need to move it above `try` block, but we might have some exceptions from RegexModifierToRegexOptions method. So we will need to put it to another try.
Not sure if we are catching exceptions "higher".

AppInspector.RulesEngine/OatExtensions/OatRegexWithIndexOperation.cs
dbalikhin: Probably, this is a new method to me, I was asking the same question myself.

AppInspector.RulesEngine/RulesVerifier.cs
dbalikhin: sure, additional optional parameter for RegexModifierToRegexOptions should work fine.

AppInspector.RulesEngine/RulesVerifier.cs
gfs: I don't think `RegexModifierToRegexOptions` has any cases where it will throw, so it should be safe to move out of the try.

AppInspector.RulesEngine/OatExtensions/OatRegexWithIndexOperation.cs
gfs: Yes, I just added this yesterday so understandable you may have missed it in the merge. I was hoping to make it a little easier to expand the modifiers available with the new method.

AppInspector.RulesEngine/RulesVerifier.cs
```diff
@@ -148,12 +148,27 @@ public RuleStatus CheckIntegrity(ConvertedOatRule convertedOatRule)
...
+                    }
+#endif
 
                     _ = new Regex(searchPattern.Pattern, regexOpts);
                 }
+                catch (NotSupportedException e)
+                {
+                    // Regex requires backtracking, try without non-backtracking
+                    _logger?.LogWarning(MsgHelp.GetString(MsgHelp.ID.VERIFY_RULES_REGEX_FAIL), rule.Id ?? "", searchPattern.Pattern ?? "", e.Message);
+                    errors.Add(MsgHelp.FormatString(MsgHelp.ID.VERIFY_RULES_REGEX_FAIL, rule.Id ?? "", searchPattern.Pattern ?? "", e.Message));
```
gfs, line 168: Can't add this message to the errors. Any messages in the errors indicate validation failure. The Validation object doesn't currently have a field for informative messages like warnings, Errors are always assumed to be a validation failure.

gfs: Actually this might not be true. Double checking the `Verified` field of the `RuleStatus` object it only checks the `OatIssues` field not the errors field.

gfs: https://github.com/microsoft/ApplicationInspector/blob/c187a93438406c47be469285d731fca950d6d8df/AppInspector.RulesEngine/RuleStatus.cs#L11

AppInspector.RulesEngine/RulesVerifier.cs
dbalikhin: Haha, you are right. For some reason I read the code below the actual method in the Utils.cs file. Will change!

AppInspector/Commands/PackRulesCommand.cs
```diff
@@ -20,6 +20,8 @@ public class PackRulesOptions
     public bool DisableRequireUniqueIds { get; set; }
     public bool RequireMustMatch { get; set; }
     public bool RequireMustNotMatch { get; set; }
+
+    public bool EnableNonBacktrackingRegex { get; set; }
```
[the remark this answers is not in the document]
dbalikhin: We do Rules Verification as a part of PackRules.  Rules Verification includes NB.
But I think you are right, and it should not be packed.

AppInspector.RulesEngine/RulesVerifier.cs
dbalikhin: Just log then.

AppInspector.RulesEngine/Ruleset.cs
```diff
@@ -14,7 +14,9 @@ public class RuleSet : TypedRuleSet<Rule>
     ///     Create a ruleset using the given (optional) logger.
     /// </summary>
     /// <param name="loggerFactory"></param>
-    public RuleSet(ILoggerFactory? loggerFactory = null) : base(loggerFactory)
+    public RuleSet(ILoggerFactory? loggerFactory = null, bool enableNonBacktrackingRegex = false) 
+        : base(loggerFactory)
     {
+        EnableNonBacktrackingRegex = enableNonBacktrackingRegex;
```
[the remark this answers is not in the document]
dbalikhin: I'm going to make it public for now if you don't mind. You can refactor later, but we won't break an API contract.

AppInspector.Common/Utils.cs
```diff
@@ -14,10 +14,14 @@ public static class Utils
...
-    public static RegexOptions RegexModifierToRegexOptions(IList<string> modifiers)
+    public static RegexOptions RegexModifierToRegexOptions(IList<string> modifiers, bool enableNonbacktrackingRegex = false)
     {
         RegexOptions opts = new();
         opts |= RegexOptions.Compiled;
+#if NET7_0_OR_GREATER
+        if (enableNonbacktrackingRegex)
+            opts |= RegexOptions.NonBacktracking;
+#endif
         foreach (var modifier in modifiers)
```
gfs, line 25: I thinl this switch list is missing two cases now that should be handled. `b` in the case that non backtrack is set and `nb` in either case. The `b` can be set with the `&= ~` technique and the nb with `|=`. Specifically for cases where it's set individually in the rule rather than by parameter.

AppInspector.RulesEngine/AbstractRuleSet.cs
```diff
@@ -77,7 +78,14 @@ public IEnumerable<ConvertedOatRule> GetUniversalRules()
         var clauses = new List<Clause>();
         var clauseNumber = 0;
         var expression = new StringBuilder("(");
+
         foreach (var pattern in rule.Patterns)
+        {
+            if (EnableNonBacktrackingRegex && !pattern.Modifiers.Contains("b") && !pattern.Modifiers.Contains("nb"))
```
gfs, line 84: This is an interesting edge case we probably want to some how catch in the validator. B and nb are exclusive options.

gfs:
Let me know when you're ready for me to do another review.

AppInspector.RulesEngine/AbstractRuleSet.cs
gfs: I also wonder, if `nb` is being added here, do we also need the parameter to `RegexModifierToRegexOptions` or is this nb enough to flow through?

AppInspector.RulesEngine/OatExtensions/OatRegexWithIndexOperation.cs
```diff
@@ -83,7 +83,7 @@ private static IEnumerable<Violation> RegexWithIndexValidationDelegate(CST.OAT.R
         if (state1 is TextContainer tc && clause is OatRegexWithIndexClause src &&
             clause.Data is List<string> RegexList && RegexList.Count > 0)
         {
-            RegexOptions regexOpts = Common.Utils.RegexModifierToRegexOptions(src.Arguments);
+            RegexOptions regexOpts = Common.Utils.RegexModifierToRegexOptions(src.Arguments, enableNonbacktrackingRegex: src.Arguments.Contains("nb"));
```
gfs, line 86: I think this is not necessary, since `nb` is added to the arguments when the pattern is added to the abstract rule set - as long as the `nb` and `b` cases are added to `RegexModifierToRegexOptions`

AppInspector.RulesEngine/RulesVerifier.cs
```diff
@@ -142,18 +142,27 @@ public RuleStatus CheckIntegrity(ConvertedOatRule convertedOatRule)
...
+                                        
                     _ = new Regex(searchPattern.Pattern, regexOpts);
                 }
+                catch (NotSupportedException e)
+                {
+                    // Regex requires backtracking, try without non-backtracking
+                    _logger?.LogWarning(MsgHelp.GetString(MsgHelp.ID.VERIFY_RULES_REGEX_FAIL), rule.Id ?? "", searchPattern.Pattern ?? "", e.Message);
+#if NET7_0_OR_GREATER
+                    // remove non-backtracking flag by applying XOR
+                    regexOpts ^= RegexOptions.NonBacktracking;
```
gfs, line 161:
```suggestion
                    regexOpts &= ~RegexOptions.NonBacktracking;
```

I think its a bit clearer to just not the specific bit rather than xoring.

gfs (commented):
Overall looking pretty good I think. A few more comments. Before merging will need a few tests still.

AppInspector.RulesEngine/AbstractRuleSet.cs
gfs: Another comment on this line. `b` and `nb` should be checked case insensitively to match implementation in `RegexModifierToRegexOptions`.

Perhaps

`pattern.Modifiers.Any(modifier => modifier.Equals("b", StringComparison.CurrentCultureIgnoreCase)`

gfs:
@dbalikhin Thanks for your work on this so far! I think its getting pretty close. Just let me know if you'd like me to help make the changes I proposed above, or if you have any concerns about them.

dbalikhin:
hey @gfs, I will try to complete it on this weekend. If for some reason it is not done (well, family stuff), feel free to make changes.

AppInspector.Common/Utils.cs
dbalikhin: Changed to `&= ~`

AppInspector.RulesEngine/AbstractRuleSet.cs
dbalikhin: I rewrote this piece. I'm using string comparison InvariantCultureIgnoreCase similar to all cases in this file. Just keeping the same style.

AppInspector.RulesEngine/RulesVerifier.cs
dbalikhin: Yeah, changed the logic, a bit of edge case, but your option is better.

dbalikhin:
I added some validation logic to AbstractRuleSet. It should handle mutually exclusive `b` and `nb` now.
I will review the tests project and create some tests.

dbalikhin:
Hm, it does look like we don't have rule validation if it comes from a default ruleset.
Should we modify StringToRegex to catch this edge case or ruleverify command needs to be used every time before changing a default ruleset?
```csharp 
   private Regex? StringToRegex(string built, RegexOptions regexOptions)
    {
        if (!RegexCache.ContainsKey((built, regexOptions)))
        {
            try
            {
                RegexCache.TryAdd((built, regexOptions), new Regex(built, regexOptions));
            }
            catch (NotSupportedException)
            {
                _logger.LogWarning("Provided regex {Regex} was not supported and could not be used", built);
#if NET7_0_OR_GREATER
                regexOptions &= ~RegexOptions.NonBacktracking;
                RegexCache.TryAdd((built, regexOptions), new Regex(built, regexOptions));
#endif
```

Trying to write some test cases and got exceptions in StringToRegex.

gfs:
When changing the default rule set I either run the verify command against the default rules or use the unit test that validates default rules (using the verify command under the hood).

Since we have the unit test in CI it seemed redundant to me to have analyze also validate the rules - as they are embedded in the binary so should not have changed from the build step to the publication.

gfs:
When changing the default rule set I either run the verify command against the default rules or use the unit test that validates default rules (using the verify command under the hood).

Since we have the unit test in CI it seemed redundant to me to have analyze also validate the rules - as they are embedded in the binary so should not have changed from the build step to the publication. Verify does definitely have to be run every time there's a change to the default rules. 

As for the second question, I'm not totally sure what sequence of events is occurring you're having trouble with, what is the exception thrown?

gfs:
I reread the above, and think I understand what you mean now. I don't think the absence of rule verification of the default rules on each run should be an issue, since they are verified in the build step via CI and are embedded in the binary.

gfs:
@dbalikhin I added a slight modification to your proposed fix for `StringToRegex` and fixed up the multiple places where regexes were being created. I can revisit this next week to add any needed tests, but i think overall looks good and should be able to merge next week. Thanks again!

gfs (commented):
LGTM, pending adding a few more test cases, I can address next week and aim to merge next week as well.

Some tests that would be good to add:
Test the Utils.StringToRegex method directly
Test the behavior of a ruleset with a rule not compatible with non backtracking with nonbacktracking set
Test that the rule verifier warns about having both `b` and `nb`, or if `nb` is set and the rule is not `nb` compatible.

AppInspector.Common/Utils.cs
```diff
@@ -9,6 +9,50 @@ namespace Microsoft.ApplicationInspector.Common;
...
+#if NET7_0_OR_GREATER
+            // Its possible that this regex is not compatible with the non-backtracking engine
+            // Try constructing it without NonBackTracking
+            regexOptions &= ~RegexOptions.NonBacktracking;
+            try
+            {
+                return new Regex(built, regexOptions);
+            }
+            catch (Exception)
+            {
```
gfs, line 36: Possibly will want to add an optional logger parameter to this to get information about what exceptions were found for the rule verifier.

AppInspector/Commands/VerifyRulesCommand.cs
```diff
@@ -102,13 +103,14 @@ public VerifyRulesResult GetResult()
...
+                LanguageSpecs = Languages.FromConfigurationFiles(_loggerFactory, _options.CustomCommentsPath, _options.CustomLanguagesPath),
+                EnableNonBacktrackingRegex = _options.EnableNonBacktrackingRegex,
             };
             RulesVerifier verifier = new(options);
             verifyRulesResult.ResultCode = VerifyRulesResult.ExitCode.Verified;
 
-            RuleSet? ruleSet = new(_loggerFactory);
+            RuleSet? ruleSet = new(_loggerFactory) { EnableNonBacktrackingRegex = options.EnableNonBacktrackingRegex };
+
             if (_options.VerifyDefaultRules)
```
gfs, line 114: Should add a specific check in here for incompatible `nb` and `b` both specified - and cases with Regex with `nb` explicitly set that cannot be constructed as `nb` (but not those which cannot be constructed with `nb` but only because the `EnableNonBacktrackingRegex` option was set.

dbalikhin:
I added some tests, feel free to modify anything. 

> "the rule verifier warns about having both b and nb, or if nb is set and the rule is not nb compatible."
Probably rule conversion will handle these edge cases. I think it works just fine with the extra logic around RegexCache that you added. Check my tests, they aren't for the verify command, but a generated ruleset should work after.

NB enabled with b and nb: b overwrites nb
NB disabled with b and nb: nb overwrites b
NB is enabled/set, but the rule is not compatible, nb removed from this regex.

gfs (approved):
LGTM.

gfs:
Awesome contribution @dbalikhin! Definitely appreciate your work here!  I plan to merge this this afternoon.

PR #540, opened 2023-05-04
Provide filepath to the TextContainer for logging messages

gfs:
The TextContainer provides logging messages the first time we attempt to parse a document as xml/yml/json and fail, but this message crucially did not contain the filename. This adds the filename to those error messages to clarify what they refer to.

### DIFF
--- a/AppInspector.RulesEngine/TextContainer.cs
+++ b/AppInspector.RulesEngine/TextContainer.cs
@@ -37,8 +37,9 @@ public class TextContainer
     private object _xpathLock = new();
 
     private bool _triedToConstructYmlDocument;
-    private YamlStream _ymlDocument;
+    private YamlStream? _ymlDocument;
     private object _yamlLock = new();
+    private readonly string? _filePath;
 
     /// <summary>
     ///     Creates new instance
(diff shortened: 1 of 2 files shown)


AppInspector.RulesEngine/TextContainer.cs
```diff
@@ -383,7 +387,7 @@ public bool IsCommented(int index)
     /// <param name="scopes"> The scopes to check </param>
     /// <param name="boundary"> Boundary in the text </param>
     /// <returns> True if boundary is in a provided scope </returns>
-    public bool ScopeMatch(IEnumerable<PatternScope> scopes, Boundary boundary)
+    public bool ScopeMatch(IList<PatternScope> scopes, Boundary boundary)
```
daalcant, line 390: Curious about the reason for the change from IEnumerable to IList here

daalcant (approved):

AppInspector.RulesEngine/TextContainer.cs
gfs: It's more correct to what this intended to do. This api usually takes an array. But if someone were to provide an ienumerable that is not a list it would be enumerated multiple times in this method. We don't gain anything from the extra flexibility of enumerable so list serves the purpose here better. But IList allows either an array or a List.

PR #546, opened 2023-06-07
Add Support for Xpath Queries with Namespaces

gfs:
Fixes an issue calculating the correct set of rules for a filename in some cases.
Fixes XPATH parsing with self closing tags
Adds comment style support for xml language.
Adds test for XML query with namespace.
Adds support for excluding files from a rule based on regular expressions.


atrompler (approved):
LGTM

PR #558, opened 2023-08-02
Adds numbering to issue results

### LINKED ISSUE #556: Number the Source File Listing results in the output.html file
It takes my eyes a moment to figure out where individual links end and the next links begin.  It would also be helpful to be able to say which link one is referring to using a number, such as "link number three". 

Suggest numbering each of the the Source File Listing results as shown in the image I've added to this Issue. Doesn't need to be red font or super big font. 😄 

image

gfs:
Fix #556.

Now display a count for the issue results when drilling down into a finding in the key features tab.
Also should fix an issue with the line numbers displayed on the sidebar for findings in the first few lines of a file.

Example:
image

### DIFF
--- a/AppInspector.CLI/html/resources/js/appinspector.js
+++ b/AppInspector.CLI/html/resources/js/appinspector.js
@@ -41,7 +41,9 @@
         const endLocationLine = $(e.target).data('endLocationLine');
         const editor = ace.edit("editor");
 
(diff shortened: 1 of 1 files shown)


daalcant (approved):

cqueern:
This looks great, thank you @gfs ! Big win!

image

PR #559, opened 2023-08-29
JSON rules cleanup

gvb84:
The JSON rules are not always consistent. There are rules with unused fields, rules with fields being encountered twice or just completely wrong fields that don't have any meaning in the context of ApplicationInspector. This is not a big issue let alone a issue at all for appinspect itself. However if one is parsing the JSON rules with a stricter JSON parser which has more validation enabled this becomes an issue.

### DIFF
--- a/AppInspector/rules/default/cloud_services/cloud_hosting.json
+++ b/AppInspector/rules/default/cloud_services/cloud_hosting.json
@@ -197,7 +197,6 @@
       "CloudServices.Hosting.DigitalOcean"
     ],
     "severity": "moderate",
-    "rule_info": "",
     "patterns": [
       {
         "pattern": "digitalocean.com",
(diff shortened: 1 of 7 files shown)


gvb84:
> `@microsoft-github-policy-service agree [company="{your company}"]`

@microsoft-github-policy-service agree [company="Anvil Secure"]

gfs:
@gvb84 Thank you for the PR! Definitely appreciate the cleanup efforts.

I have been thinking it may be worth having a format schema definition for the json format, opened an issue to track that #560.

gfs:
Tests are succeeding locally, but also must run as a PR gate. Need to investigate why the pipeline isn't triggering here. @daalcant

gfs (approved):

daalcant:
Updated the PR pipeline fork triggers, but that didn't help. Kicked off a manual run to (hopefully) get this unblocked.

PR #584, opened 2024-04-29
Fix off by one error in GetPrefixLocation recursion

gfs:


danfiedler-msft (approved):
LGTM but this also seems like a good opportunity to add a test for this condition.

gfs:
Will add a test case before merging.

scovetta (approved):

PR #609, opened 2025-05-09
Fix the privacy statement

emmanuel-ferdman:
# PR Summary
This small PR fixes the Microsoft Privacy Statement reference added in #606 PR.


gfs (approved):
Thanks for the fix!

PR #619, opened 2025-07-17
change release pipeline to use same templates as PR pipeline

danfiedler-msft:


gfs (approved):

danfiedler-msft:
Unfortunately, the ADO repository resource only supports repositories in the same project or same organization. It cannot be used for repositories in other organizations. See [ADO docs](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/resources?view=azure-devops&tabs=schema#repository-resource-types)

PR #642, opened 2026-02-18
Fix JSON serialization for Result-derived classes

### LINKED ISSUE #641: Bug: `exporttags` with `--output-file-format json` outputs only appVersion, no tags
**Describe the bug**
The `exporttags` command with `--output-file-format json` produces a JSON file containing only the `appVersion` field. The actual tags are completely missing from the output. The text format works correctly.

**To Reproduce**
1. Run `appinspector exporttags --output-file-format json --output-file-path .\tags.json`
2. Open `tags.json`
3. Observe the file contains only:
   ```json
   {
     "appVersion": "Microsoft Application Inspector 1.9.55+0b8c96ee0b"
   }
   ```
4. No tags are present in the output

**Expected behavior**
The JSON output should contain all exported tags in a structured format, equivalent to the text format output which correctly lists all tags.

**Screenshots**

Image

**Operating Environment (please complete the following information):**
 - Application Inspector Version: 1.9.55 (latest on NuGet as of Feb 2026)
 - OS: Windows
 - Browser: N/A
 - Browser Version: N/A

**Additional context**
Using `--output-file-format text` works correctly as a workaround and outputs all tags:
```
appinspector exporttags --output-file-format text --output-file-path .\tags.txt
```
Installed via `dotnet tool install -g Microsoft.CST.ApplicationInspector.CLI`.

felickz:
* Fix exporttags JSON output bug by using runtime type for serialization
* Improve test to exercise actual JsonWriter code path
  - Modified ExportJsonSerialization test to use JsonWriter.WriteResults
  - Added InternalsVisibleTo attribute to expose JsonWriter to tests
  - Test validates production code path instead of direct serialization
  
  
  Fixes: #641

### DIFF
--- a/AppInspector.CLI/Properties/AssemblyInfo.cs
+++ b/AppInspector.CLI/Properties/AssemblyInfo.cs
@@ -1,0 +1,3 @@
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("AppInspector.Tests")]
--- a/AppInspector.CLI/Writers/JsonWriter.cs
+++ b/AppInspector.CLI/Writers/JsonWriter.cs
@@ -39,7 +39,7 @@ internal class JsonWriter : CommandResultsWriter
                 case TagDiffResult:
                 case ExportTagsResult:
                 case VerifyRulesResult:
-                    JsonSerializer.Serialize(StreamWriter.BaseStream, result, options);
+                    JsonSerializer.Serialize(StreamWriter.BaseStream, result, result.GetType(), options);
                     break;
                 case PackRulesResult prr:
                     JsonSerializer.Serialize(StreamWriter.BaseStream, prr.Rules, options);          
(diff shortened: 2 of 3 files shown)


gfs (approved):

gfs (approved):
Will make follow up change in another PR.

gfs:
Thanks @felickz — the diagnosis and fix here are correct, and I've carried your commit over unchanged (authorship and `Co-authored-by` trailers intact) in #648.

Closing this one only because it can't be merged from where it sits: the fork is organization-owned, so GitHub doesn't offer the "Allow edits by maintainers" option, and the required Azure pipelines don't run against forks — hence the `/azp run` attempts above returning "No pipelines are associated with this pull request." There was no path to get this green in place.

Two things came out of review that #648 adds on top of your commit:

1. **`tagdiff -f json` was still broken after the fix.** The switch case you corrected also covers `TagDiffResult`, but `TagDiffResult.TagDiffList` is declared as a public *field*, and System.Text.Json skips fields unless `IncludeFields` is set. Output was still `{ "resultCode": 1, "appVersion": "..." }`. Changed it to a property.

2. **The `InternalsVisibleTo` turned out to be avoidable.** `ResultsWriter.Write` is public and is what `Program.cs` actually calls, so the test can go through it to a temp file and get writer-selection coverage for free. The tests moved into a new `TestJsonWriter` covering all three commands on that shared path.

Appreciate you tracking this down and filing #641 with the repro.

PR #648, opened 2026-08-01
Fix JSON serialization for Result-derived classes

### LINKED ISSUE #641: Bug: `exporttags` with `--output-file-format json` outputs only appVersion, no tags

gfs:
Fixes #641. Supersedes #642.

This carries @felickz's commit from #642 unchanged, plus two follow-up commits from review. #642 can't be merged from its fork — the fork is organization-owned (so `maintainer_can_modify` is unavailable) and the required Azure pipelines don't run on forks, leaving the PR permanently blocked. Rebasing it here is the only way to land it. Original authorship and `Co-authored-by` trailers are preserved on `e561b9b`.

## Commits

**`e561b9b` — Fix JSON serialization for Result-derived classes** (@felickz, unchanged from #642)

`JsonSerializer.Serialize(stream, result, options)` bound `TValue` to the static type `Result`, so only `appVersion` was emitted for `exporttags`, `tagdiff`, and `verifyrules`. Passing `result.GetType()` selects the runtime converter.

**`be00ce7` — Serialize `tagDiffList` by making `TagDiffList` a property**

The runtime-type fix alone doesn't repair `tagdiff --output-file-format json`. `TagDiffResult.TagDiffList` was declared as a public *field*, and System.Text.Json skips public fields unless `IncludeFields` is set, so the output was still:

```json
{ "resultCode": 1, "appVersion": "..." }
```

Declaring it as a property matches every other `Result`-derived class and is narrower than enabling `IncludeFields` globally, which would start emitting any other public field on any serialized type.

**`3fd93a9` — Cover the json writers through the public `ResultsWriter` path**

Consolidates the regression tests into `TestJsonWriter`, following the existing `TestMarkdownWriter` pattern, covering all three commands that share the `JsonWriter` path rather than `exporttags` alone.

The tests now go through `ResultsWriter.Write` — what `Program.cs` calls — so writer selection is exercised alongside serialization. That's reachable without widening the CLI assembly's internal surface, so the `InternalsVisibleTo` added in #642 for direct `JsonWriter` construction is removed.

Assertions compare the written document against the in-memory result instead of substring-matching raw JSON, which would have passed on malformed output. `VerifyRulesResult` is asserted via `JsonDocument` because `RuleStatus` exposes a computed `Verified` and OAT-owned `Violation` members that don't round-trip.

## Verification

- Solution builds across net8.0 / net9.0 / net10.0.
- Full suite: 350/350 pass (348 baseline + 3 new − 1 replaced).
- Reverting the `JsonWriter` line fails all three new tests; reverting the field-to-property change fails `TagDiffWritesTagDiffList` specifically.

## Not addressed

`DefaultIgnoreCondition = WhenWritingDefault` drops `resultCode` from JSON output whenever it's `0` (`Success` / `Verified` / `TestPassed`). Pre-existing and unrelated to the base-type bug, so left alone here.

### DIFF
--- a/AppInspector.CLI/Properties/AssemblyInfo.cs
+++ b/AppInspector.CLI/Properties/AssemblyInfo.cs
@@ -1,0 +1,3 @@
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("AppInspector.Tests")]
--- a/AppInspector.CLI/Properties/AssemblyInfo.cs
+++ b/AppInspector.CLI/Properties/AssemblyInfo.cs
@@ -1,3 +1,0 @@
-using System.Runtime.CompilerServices;
-
-[assembly: InternalsVisibleTo("AppInspector.Tests")]
--- a/AppInspector.CLI/Writers/JsonWriter.cs
+++ b/AppInspector.CLI/Writers/JsonWriter.cs
@@ -39,7 +39,7 @@ internal class JsonWriter : CommandResultsWriter
                 case TagDiffResult:
                 case ExportTagsResult:
                 case VerifyRulesResult:
-                    JsonSerializer.Serialize(StreamWriter.BaseStream, result, options);
+                    JsonSerializer.Serialize(StreamWriter.BaseStream, result, result.GetType(), options);
                     break;
                 case PackRulesResult prr:
                     JsonSerializer.Serialize(StreamWriter.BaseStream, prr.Rules, options);          
--- a/AppInspector.Tests/Commands/TestExportTagsCmd.cs
+++ b/AppInspector.Tests/Commands/TestExportTagsCmd.cs
@@ -59,4 +59,41 @@ public class TestExportTagsCmd
         };
         Assert.Throws<OpException>(() => new ExportTagsCommand(options));
     }
+
+    [Fact]
+    public void ExportJsonSerialization()
+    {
+        ExportTagsOptions options = new()
+        {
+            IgnoreDefaultRules = true,
+            CustomRulesPath = testRulesPath
+        };
+        ExportTagsCommand command = new(options, factory);
+        var result = command.GetResult();
+        
+        // Test JSON serialization via the production JsonWriter to ensure tags are included
+        using var memoryStream = new MemoryStream();
+        using var streamWriter = new StreamWriter(memoryStream);
+        
+        // Use the JsonWriter that production code uses
+        var jsonWriter = new Microsoft.ApplicationInspector.CLI.Writers.JsonWriter(streamWriter, factory);
+        var cliOptions = new Microsoft.ApplicationInspector.CLI.CLIExportTagsCmdOptions
+        {
+            IgnoreDefaultRules = true,
+            CustomRulesPath = testRulesPath
+        };
+        
+        jsonWriter.WriteResults(result, cliOptions, autoClose: false);
+        streamWriter.Flush();
+        
+        memoryStream.Position = 0;
+        using var reader = new StreamReader(memoryStream);
+        string json = reader.ReadToEnd();
+        
+        // Verify tags are present in JSON output from JsonWriter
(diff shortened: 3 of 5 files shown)


gfs:
## SDL Security Review

> _Automated output from the **COMPASS SDL Security Review** skill, run against the diff for this PR. This is tool output, not a personal review — findings are advisory and reflect the skill's ruleset, not a human sign-off._

**Scope:** 4 files, +144/−3.

| File | Change |
|---|---|
| `AppInspector.CLI/Writers/JsonWriter.cs` | Pass `result.GetType()` to `JsonSerializer.Serialize` |
| `AppInspector/Commands/TagDiffCommand.cs` | `TagDiffList` field → property |
| `AppInspector.Tests/Commands/TestJsonWriter.cs` | New test file |
| `AppInspector.Tests/Commands/TestExportTagsCmd.cs` | Trailing newline |

### 🔴 Blocking

None.

### 🟡 Warning

None.

### 🟢 Informational

**`AppInspector.CLI/Writers/JsonWriter.cs`** — the fix intentionally widens the serialized surface:

```csharp
JsonSerializer.Serialize(StreamWriter.BaseStream, result, result.GetType(), options);
```

`verifyrules --output-file-format json` now emits `RuleStatusList`, including `Errors`, `OatIssues` (OAT `Violation`) and `SchemaValidationErrors`, where output previously collapsed to `appVersion` alone. The producers (`RulesVerifier.cs:115`, `:135`, `:434-438`) carry rule IDs, tag names and schema diagnostics from the operator's own ruleset — no secrets, credentials or PII, and no trust boundary is crossed, since this is a local CLI writing to a user-specified path. Not a finding; noted only because reports published as CI artifacts will be materially larger than before.

### Checks cleared

- **Deserialization** — the new `JsonSerializer.Deserialize<T>` calls are test-only, operate on data the test just wrote, and use plain System.Text.Json with no `JsonDerivedType` or polymorphic resolver, so they are not a `TypeNameHandling`-style gadget surface. The production change is serialization only.
- **Attack surface** — `be00ce7` removes the `InternalsVisibleTo` that `e561b9b` added to `AppInspector.CLI/Properties/AssemblyInfo.cs`, net zero in the final diff. Tests reach the writer through the public `ResultsWriter.Write` path instead, which is the narrower of the two options.
- **Field → property on `TagDiffList`** — adds a public setter on a local CLI result object. No model-binding or mass-assignment boundary, and the constructor still initializes it.
- **Path handling (tests)** — rule and sample paths are `Path.Combine` over constants, not user input. The temp file is `Path.GetTempPath()` + `Guid.NewGuid()`, write-only, deleted in a `finally`.
- **Secrets, cryptography, injection, authn/authz, memory safety** — no relevant code in this diff.
- No `.github/`, pipeline, `nuget.config` or dependency changes.

**Result: APPROVE** — no blocking SDL issues found.

chanel-y (approved):

